PR #161, opened 2018-06-22
Issue #157 feat: saving ID in a new internal property

steotia:
@anandp504 I would like you to review the PR.

In this PR,
- we can now use remote Neo4J
- the entities have an internal ID but they can be addressed by both the external and internal ID
- All V() type queries are replaced


indrajra:
I notice this PR brought in the necessary change of encapsulating the id field generation. But this missed achieving robust scalability on remote databases. Closing this PR. Thanks!

PR #162, opened 2018-06-26
Deployment configuration

jyotsna-tarento:
Build and deployment scripts


Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM frolvlad/alpine-oraclejdk8:slim
+COPY ./java/registry/target/registry.jar /registry.jar
```
anandp504: Please check if you can copy the file under /home/opensaber directory. Also, you need to provide a mapping for the Neo4J database directory and logs directory as we have done in the docker-compose file

metadata.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# Return service metadata
+echo '{"name":"registry","version":"1.0.0","org":"opensaber","hubuser":""}'
```
anandp504: Please check if you can update the version using pom.xml

jyotsna-tarento:
1. Fetching properties from pom.xml and replacing placeholders in scripts have been implemented.
2. Copying the jar to the path mentioned has also been implemented.
3. Volumes have not been included in docker file.

PR #169, opened 2018-08-30
Issue #OS-43:changes for persistence of digital signatures

jyotsna-tarento:
This PR includes the following:

1.Changes in Validation schema to support digital signature for add entity operation.
2. Changes in RDF to Graph conversion to incorporate changes as discussed during design.
3. Changes in Dao layer to allow storing of signatures.
4. Refactoring some of the common methods used by create and update operations in Dao layer as there was repetitive code which was also required  by Digital Signature feature.

### DIFF
--- a/java/converters/rdf2Graph/src/main/java/io/opensaber/utils/converters/RDF2Graph.java
+++ b/java/converters/rdf2Graph/src/main/java/io/opensaber/utils/converters/RDF2Graph.java
@@ -63,9 +63,7 @@ public final class RDF2Graph
 		Vertex s = getExistingVertexOrAdd(subjectValue.toString(), graph);
 		Vertex subjectVertex = s;
 		if(property.toString().equalsIgnoreCase(context+Constants.SIGNATURES)){
-			Vertex proofVertex = getExistingSignatureVertexOrAdd(context, graph);
-			proofVertex.addEdge(context+Constants.SIGNATURE_OF, s);
-			subjectVertex = proofVertex;
+			subjectVertex = getExistingSignatureVertexOrAdd(s, context, graph);
 		}
 		if (objectValue instanceof Literal) {
 			Literal literal = (Literal)objectValue;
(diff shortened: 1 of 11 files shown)


java/registry/src/main/java/io/opensaber/registry/dao/impl/RegistryDaoImpl.java
```diff
@@ -234,91 +234,161 @@ private String addOrUpdateVerticesAndEdges(GraphTraversalSource dbTraversalSourc
...
         }
         for (Pair<Vertex, Vertex> pv : parsedVertices) {
-            addOrUpdateVertexAndEdge(pv.getValue0(), pv.getValue1(), dbGraph, methodOrigin, encDecPropertyBuilder);
+        	addOrUpdateVertexAndEdge(pv.getValue0(), pv.getValue1(), dbGraph, methodOrigin, encDecPropertyBuilder, Direction.OUT);
         }
 
 
+    }
+
+    private Stack<Pair<Vertex, Vertex>> addOrUpdateVertexAndEdgeIteratively(Iterator<Edge> edges, Iterator<Edge> edgeList, List<Edge> dbEdgesForVertex,
```
indrajra: no methods like this one please...try calling it differently. Additionally split this to Add separately and Update separately. Remember, we agreed to get rid of methodOrigin.

java/registry/src/main/java/io/opensaber/registry/dao/impl/RegistryDaoImpl.java
```diff
@@ -234,91 +234,161 @@ private String addOrUpdateVerticesAndEdges(GraphTraversalSource dbTraversalSourc
...
+     * @param parsedVertices
+     * @param dbVertex
+     * @param e
+     * @param edgeVertexMatchList
+     * @param edgeVertexAlreadyExists
+     * @return
+     * @throws AuditFailedException
+     * @throws EncryptionException
+     */
+    private Stack<Pair<Vertex, Vertex>> addEdgeForAVertex(Vertex ver, Vertex newV, GraphTraversalSource dbGraph, String methodOrigin,
```
indrajra: what has parsedVertices got to do with adding an edge to a vertex? I think this method takes an increased number of params and must be trimmed.

java/registry/src/main/java/io/opensaber/registry/dao/impl/RegistryDaoImpl.java
jyotsna-tarento: Yes, we agreed on getting rid of methodOrigin, but it's better to take it up as a separate task as it's used across the entire Dao layer. But the reason why the methods were refactored along with this was that, during adding of signature, changing this existing code used by add and update methods, was getting complicated. So that's why splitting into smaller methods seemed better right now, to simplify and group common changes.

java/converters/rdf2Graph/src/main/java/io/opensaber/utils/converters/RDF2Graph.java
```diff
@@ -73,19 +82,29 @@ private static void updateGraph(Value subjectValue, IRI property, Value objectVa
...
+			subjectVertex.addEdge(property.toString(), o);
 		} else if (objectValue instanceof BNode) {
 			BNode objectBNode = (BNode)objectValue;
 			Vertex o = getExistingVertexOrAdd(objectBNode.toString(), graph);		
-			s.addEdge(property.toString(), o);
+			subjectVertex.addEdge(property.toString(), o);
+		}
+	}
+	
+	private static Vertex getExistingSignatureVertexOrAdd(String context, Graph graph){
```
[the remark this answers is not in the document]
jyotsna-tarento: This class contains methods which converts the RDF model to a Graph Object. It is checking if a vertex exists in the in-memory graph and not in the database. This graph is created in this layer itself, during each iteration of the statements present in the RDF model. So that's why it is having this logic.

PR #171, opened 2018-09-11
Issue: OS 56 Factory impls for response content based on request Mime…

pritha-tarento:
Implementation for response content based on request Mime type.

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
+++ b/java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
@@ -5,16 +5,24 @@ import java.util.HashMap;
 import java.util.Map;
 import io.opensaber.pojos.*;
 import io.opensaber.registry.middleware.util.Constants;
+import io.opensaber.registry.response.content.ResponseContent;
+import io.opensaber.registry.response.content.ResponseContentFactory;
 import io.opensaber.registry.util.JSONUtil;
+
 import org.apache.jena.rdf.model.Model;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
+import org.springframework.util.MimeType;
+import org.springframework.util.MimeTypeUtils;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestAttribute;
+import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
@@ -50,6 +58,9 @@ public class RegistryController {
 	
 	@Value("${audit.enabled}")
 	private boolean auditEnabled;
+	
+	@Autowired
+	ResponseContentFactory responseContentFactory;
 
 	@Autowired
 	private OpenSaberInstrumentation watch;
@@ -70,7 +81,7 @@ public class RegistryController {
 			response.setResult(result);
 			responseParams.setStatus(Response.Status.SUCCESSFUL);
 			watch.stop("RegistryController.addToExistingEntity");
-			logger.debug("RegistryController : Entity with label {} added !", label);
+			logger.info("RegistryController : Entity with label {} added !", label);
 		} catch (DuplicateRecordException | EntityCreationException e) {
 			logger.error("DuplicateRecordException|EntityCreationException in controller while adding entity !",e);
 			response.setResult(result);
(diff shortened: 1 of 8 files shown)


java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
```diff
@@ -86,21 +97,29 @@
...
 			org.eclipse.rdf4j.model.Model entityModel = registryService.getEntityById(entityId);
 			logger.debug("FETCHED: " + entityModel);
-			String jenaJSON = registryService.frameEntity(entityModel);
-			response.setResult(gson.fromJson(jenaJSON, mapType));
+			/*String jenaJSON = registryService.frameEntity(entityModel);
+			response.setResult(gson.fromJson(jenaJSON, mapType));*/
+						
+			ResponseContent responseContent = responseContentFactory.getResponseContent(acceptType.toString());
+			responseContent.setModel(entityModel);
+			response.setResult(responseContent.getContent().getResult());
```
indrajra: responseContent.getContent().getResult() doesn't read well...can it be responseContent.getResult() directly?

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
```diff
@@ -86,21 +97,29 @@
...
 	@RequestMapping(value = "/read/{id}", method = RequestMethod.GET)
-	public ResponseEntity<Response> readEntity(@PathVariable("id") String id) {
+	public ResponseEntity<Response> readEntity(@PathVariable("id") String id,
+			@RequestHeader HttpHeaders headers) {
 
 		String entityId = registryContext + id;
 		ResponseParams responseParams = new ResponseParams();
 		Response response = new Response(Response.API_ID.READ, "OK", responseParams);
 
+		MediaType acceptType = headers.getAccept().iterator().next();
```
indrajra: Lets read the acceptType and invalidate the request if an un-acceptable acceptType is passed. If its empty, which means not specified then we could allow. Maybe we could have a separate function to help us.
Also, acceptTypes could be multiple in nature. Can we use @RequestHeader("accept") instead of all the http headers (just curious)?

java/registry/src/main/java/io/opensaber/registry/response/content/ResponseContent.java
```diff
@@ -0,0 +1,13 @@
+package io.opensaber.registry.response.content;
+
+import java.io.IOException;
+
+import io.opensaber.registry.exception.EntityCreationException;
+import io.opensaber.registry.exception.MultipleEntityException;
+
+public interface ResponseContent {
+	
+	public void setModel(org.eclipse.rdf4j.model.Model entityModel);// TODO: Custom model
```
indrajra: Lets not use RDF here...something else, I dont; know what it could be but leaving a thought to you. I'll update you in person - the idea is to work on Json or JsonLD in all layers, which means this must be a JSON object...today we pass around RDF and Tinkerpop in few. We want to limit this.

java/registry/src/main/java/io/opensaber/registry/response/content/ResultContent.java
```diff
@@ -0,0 +1,19 @@
+package io.opensaber.registry.response.content;
+
+import java.util.Map;
+
+public class ResultContent {
+	
+
+	private Map<String, Object> result;
```
indrajra: I dont know why should we use a Map here - I find Object is good enough and we could pass a JSONObject or whatever...in most cases, I see we just add 'response' for a want of string. Can you have your eyes on this please.

java/registry/src/main/java/io/opensaber/registry/response/content/impl/JsonldResponseContent.java
```diff
@@ -0,0 +1,222 @@
...
+	/**
+	 * TODO: for fetchAudit, needs to be generic
+	 * @param jenaEntityModel
+	 * @return
+	 * @throws IOException
+	 * @throws EntityCreationException
+	 * @throws MultipleEntityException
+	 */
+	private JsonLDWriteContext prepareJsonLDWriteContext(Model jenaEntityModel) throws IOException, EntityCreationException, MultipleEntityException{
+		JsonLDWriteContext ctx = new JsonLDWriteContext();
```
indrajra: Lets this function callers give you JsonLDWriteContext and fileString. I think this function presently does many things, asking you to trim it down.

java/middleware-commons/src/main/java/io/opensaber/registry/middleware/transform/commons/ResponseData.java
```diff
@@ -0,0 +1,14 @@
+package io.opensaber.registry.middleware.transform.commons;
```
indrajra: How different is this from Data?

java/pojos/src/main/java/io/opensaber/pojos/Response.java
```diff
@@ -46,7 +46,8 @@ public String getId() {
 	}
 
 	private String responseCode;
-	private Map<String, Object> result;	
+	private Map<String, Object> result;
```
indrajra: please merge this branch from release-1.0.0 first. These changes are already done there.

java/registry-interceptor/src/main/resources/context_mapping.json
```diff
@@ -0,0 +1,269 @@
+{
```
indrajra: I think this was needed only in the client library. We should not bring this additional baggage to the registry. The registry already has vocab and shex. This seems to be a third one and must be disallowed.

PR #179, opened 2018-09-28
OS#68 Sign and verify can happen for non-ld json documents as well

indrajra:
1. Must be able to sign and verify either values or json objects - this was not the case earlier, there was an assumption that it would always be json.
2. There is a possibility of null that was not recognized leading to build issues.

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/controller/RegistryUtilsController.java
+++ b/java/registry/src/main/java/io/opensaber/registry/controller/RegistryUtilsController.java
@@ -1,6 +1,7 @@
 package io.opensaber.registry.controller;
 
 import com.google.gson.Gson;
+import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.reflect.TypeToken;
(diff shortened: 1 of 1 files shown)


java/middleware/registry-middleware/signature-presence-validation/src/main/java/io/opensaber/registry/middleware/impl/SignaturePresenceValidator.java
```diff
@@ -109,8 +111,9 @@ private void addAttributesForShape(RDFNode firstNode, Model signatureConfigModel
 		if(!secondNode.equals(RDF.nil)){
             RDFNode thirdNode = RDFUtil.getFirstObject((Resource) secondNode, RDF.first.getURI(), signatureConfigModel);
             RDFNode fourthNode = RDFUtil.getFirstObject((Resource) thirdNode, SHAPE_PREDICATE_IRI, signatureConfigModel);
-
-            attributeList.add(fourthNode.toString());
+            if (attributeList != null) {
```
jyotsna-tarento: null check is not required for attributeList variable as it is always an empty list.

PR #182, opened 2018-10-04
Issue #OS-69 refactor:adding check for all request id in post calls

Hari-stackroute:
This changes includes the check for allowing same request id accessing the proper api.

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/config/GenericConfiguration.java
+++ b/java/registry/src/main/java/io/opensaber/registry/config/GenericConfiguration.java
@@ -248,6 +248,7 @@ public class GenericConfiguration implements WebMvcConfigurer {
 				.addPathPatterns("/add", "/update", "/search").order(orderIdx++);
 		registry.addInterceptor(rdfValidationInterceptor())
 				.addPathPatterns("/add", "/update").order(orderIdx++);
+
 		if (signatureEnabled) {
 			registry.addInterceptor(signaturePresenceValidationInterceptor())
 					.addPathPatterns("/add", "/update").order(orderIdx++);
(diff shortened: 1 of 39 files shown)


java/registry/src/main/java/io/opensaber/registry/config/GenericConfiguration.java
```diff
@@ -236,9 +248,30 @@ public Middleware rdfValidationMapper() {
...
+
+	/**
+	 * This method will process all the interceptors for each request
+	 * @param registry
+	 */
 	@Override
 	public void addInterceptors(InterceptorRegistry registry) {
 		int orderIdx = 1;
+		registry.addInterceptor(requestIdValidationInterceptor()).
+				addPathPatterns("/add", "/update", "/search","/utils/sign","/utils/verify").order(orderIdx++);
```
indrajra: use /** as the pattern to match. We don't want new developers to feel surprised - let them not add the new paths explicitly themselves.

Hari-stackroute: for read and delete no need to trigger this interceptor, since they are post calls and they wont contain body.

Hari-stackroute: For read and delete and health, this will return false and those api's wont work if we mark as /**.

indrajra: Understood. The other idea was to use the map and expand this yourselves. Otherwise, we now require 2 places to be changed - i) to insert into map ii) to fix the interceptor. Even if one is missed, it leads to un-intended behavior.

Hari-stackroute: Done the changes, please check.

PR #186, opened 2018-10-11
OS 76: JSON request & response namespace.

pritha-tarento:
The PR details:
1. Json response, removing the context-type(namespace)(OS-82)
2. Json request should not contain context-type(namespace), so adding the context-type while request transformation.(OS-82)
3. Supporting mime type for ALL(OS-73)
4. Included the refactoring of transform commons package.
5. Json-Ld Constants is put in single place. 

Testing done by postman.


java/registry-interceptor/src/main/java/io/opensaber/registry/interceptor/request/transform/JsonToLdRequestTransformer.java
```diff
@@ -61,6 +67,57 @@ private String getTypeFromNode(ObjectNode requestNode) throws JsonProcessingExce
...
+				replaceArrayNode(requestNode, entry.getKey(), arrayNode);
+
+			}
+
+		});
+	}
+
+	public static void replaceText(JsonNode parent, String fieldName, String newText) {
+		if (parent.has(fieldName)) {
+			((ObjectNode) parent).put(fieldName, newText);
```
indrajra: I have two questions:
1. Why put the cast here? This method must say, it can work only with ObjectNode - it is the caller that will have to adjust. 
2. What if the the parent node doesn't have this? Should it replace or not? The thought is why this field check happens within the replaceText function, could it happen out?

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
```diff
@@ -136,7 +135,7 @@
 	 * 
 	 * @param id
 	 * @param accept,
-	 *            only one mime type is supported.
+	 *            only one mime type is supported at a time.
```
indrajra: You said only the last one was used. That is more important to mention. Please do the test yourself from various ways before claiming.

java/registry/src/main/java/io/opensaber/registry/transformation/JsonToLdTransformer.java
```diff
@@ -66,7 +69,11 @@ private void purgedKeys(ObjectNode node) {
 			if (keysToPurge.contains(entry.getKey())) {
 				removeKeyNames.add(entry.getKey());
 			} else {
-				if (entry.getValue().isArray()) {
+				if (entry.getValue().isValueNode()) {
```
indrajra: Can you put a var for val = entry.getValue() and then start using? The if clauses are so lengthy otherwise and the gets happen multiple times.

java/registry/src/main/java/io/opensaber/registry/transformation/JsonToLdTransformer.java
```diff
@@ -79,10 +86,20 @@ private void purgedKeys(ObjectNode node) {
 		node.remove(removeKeyNames);
 	}
 
+	public static void replaceText(JsonNode parent, String fieldName, String newText) {
```
indrajra: JsonToLdRequestTransformer.java - there is already the same function in this file. Why static and code duplicacy?

java/registry-interceptor/src/main/java/io/opensaber/registry/interceptor/request/transform/JsonToLdRequestTransformer.java
pritha-tarento: Resolved.

java/registry/src/main/java/io/opensaber/registry/transformation/JsonToLdTransformer.java
pritha-tarento: Resolved

java/registry/src/main/java/io/opensaber/registry/transformation/JsonToLdTransformer.java
pritha-tarento: Resolved.

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
pritha-tarento: Resolved, the first accept type is picked.

PR #187, opened 2018-10-13
Os 72 RegistryUtils service single responsibility

pritha-tarento:
Details:
1. Added the POJOs for the Sign and Verify API
2. Refactor the register util to change for accepting the Request as @RequestAttribute. Removed using HttpServletRequest directly. Also simplified the code for sign & verify methods .

TODO:
Needs to test End to end. As it required the support in the enc-service for the same request structure.


java/registry/src/main/java/io/opensaber/registry/model/VerifyRequest.java
```diff
@@ -0,0 +1,19 @@
+package io.opensaber.registry.model;
+
+import java.util.List;
+
+public class VerifyRequest {
+	
+	private List<Verify> verify;
```
indrajra: Require name changes - must not be verify - it must be explicit - SignatureVerificationPayload or SignatureVerificationRequest - the moment we attach 'Request' suffix vocal communication to express this class becomes difficult, so I prefer the former.

indrajra:
The changes to enc-service has not been done yet nor very important in the light of other things. Closing this PR as it is too old by now. Leaving the branch to restart any efforts.

PR #188, opened 2018-10-17
Issue #OS-85 refactor:removed rdfvalidator interceptor and add in ser…

Hari-stackroute:
Removed RDFValidator interceptor
Removed middle-ware code changes
Changed service layer add  and update method signatures, so some of the regarding add and update failing. Will fix in OS-86 bug.
Now added wrapper method for addEntity for allowing test cases


java/registry/src/main/java/io/opensaber/registry/exception/RDFValidationException.java
```diff
@@ -0,0 +1,11 @@
+package io.opensaber.registry.exception;
+
+public class RDFValidationException extends Exception {
```
indrajra: I asked this base to be converted to CustomException.

PR #191, opened 2018-10-25
Os 95 Schema configurator to read json file

pritha-tarento:
Changes done:
1. Schema Loader to handle responsibility for loading create and update shex validations (OS-95)
2. Changes to test file for schema validation module.
3. Resolved Bug(OS-97)
 4. Test cases added for JSONUtil new methods.


Hari-stackroute:
Looks ok for me

PR #193, opened 2018-10-26
Os 98 Read API changed to POST method

pritha-tarento:
Changes Done
1. Read API changing from GET method type to POST method type.
2. Read API only supports JSON content type: since the request attibutes are not any LD related data.

Tested by postman
1. Read API testing is covered.

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
+++ b/java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
@@ -18,6 +18,9 @@ import io.opensaber.registry.service.SignatureService;
 import io.opensaber.registry.transformation.ResponseTransformFactory;
 
 import org.apache.jena.rdf.model.Model;
+import org.json.simple.JSONObject;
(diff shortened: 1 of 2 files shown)


java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
```diff
@@ -123,7 +126,50 @@
...
+			logger.error("RegistryController: Exception while reading entity!", e);
+			response.setResult(null);
+			responseParams.setStatus(Response.Status.UNSUCCESSFUL);
+			responseParams.setErrmsg("Ding! You encountered an error!");
+		}		
+
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
+	/*@RequestMapping(value = "/read/{id}", method = RequestMethod.GET)
```
Hari-stackroute: Ok, we can remove the commented code.

pritha-tarento: yes.

PR #195, opened 2018-10-31
OS-92 Create json schema definitions for Teacher example

indrajra:
A new folder - "registry/src/main/resources/public/_schemas/" has been added with Teacher related definitions. Since, these are under 'public' folder they are indeed shared with everyone. Change is made to allow this path GET operations to pass through without auth header.

Note that the json definitions lack any context (LD terminology). It is assumed it is in local domain/host. I believe if we need to be truly json and json-ld supported, then we may have make some adjustments to these definitions, which is not handled now (can be taken after). 

Existing (json) payload of Teacher has been tested against this schema in a separate stand-alone project. When I introduce the json schema validator, I would also add tests then.


java/registry/src/main/resources/public/_schemas/Teacher.json
```diff
@@ -0,0 +1,143 @@
...
+        "inServiceTeacherTrainingFromOthers",
+        "nonTeachingAssignmentsForAcademicCalendar",
+        "basicProficiencyLevel",
+        "disabilityType",
+        "trainedForChildrenSpecialNeeds",
+        "trainedinUseOfComputer"
+      ],
+      "properties": {
+        "signatures": {
+          "$id": "#/properties/signatures",
```
pritha-tarento: "#/propeties" is being used as a part of all id. Since it is of properties node, Do we still need? Or a JSONSchema's structure.

java/registry/src/main/resources/public/_schemas/Teacher.json
```diff
@@ -0,0 +1,143 @@
...
+        },
+        "trainedForChildrenSpecialNeeds": {
+          "$id": "#/properties/trainedForChildrenSpecialNeeds",
+          "$ref": "Common.json#/definitions/YesNoCode"
+        },
+        "trainedinUseOfComputer": {
+          "$id": "#/properties/trainedinUseOfComputer",
+          "$ref": "Common.json#/definitions/YesNoCode"
+        },
+        "___encryptedFields": {
```
pritha-tarento: Encryption node is empty?

java/registry/src/main/resources/public/_schemas/Common.json
```diff
@@ -0,0 +1,149 @@
...
+      ]
+    },
+    "Year": {
+      "type": "string",
+      "pattern": "(^[0-9]{4}$)"
+    },
+    "SocialCategory": {
+      "type": "string",
+      "enum": [
+        "SocialCategoryTypeCode-GENERAL",
```
pritha-tarento: Since the strings(of each enum) is quite lengthy, I was wondering whether it could start with Code-ABC only. As node key is saying what it is.

java/registry/src/main/resources/public/_schemas/Teacher.json
indrajra: This is a general way to uniquely identify this jsonobject. We dont use it, but @id is a way that exists to uniquely identify.

java/registry/src/main/resources/public/_schemas/Teacher.json
indrajra: oops, I didn't want to introduce this. I wanted to take care of encryption later on. Removed this key for now.

java/registry/src/main/resources/public/_schemas/Common.json
indrajra: Good point, i don't want to take work on changing all our Postman collection, documentation for this small a change. Since these are examples, let them remain as-is.

PR #201, opened 2018-11-09
OS#103 Add json based schema validation

indrajra:
- Introduced a new module “jsonvalidator” that would use everit-schema (json schema validation library). This is a sub-module under the parent “validators”.
- Two sections existed in the application.yml - Validations and validation. I like to retain validation as the name. Settings previously under validations are moved.
- Two vars introduced in the application.yml (configuration):
1. validation_type of type enum “json”, “shex”, defaulting to “shex”
2. validation_enabled of type boolean, defaulting to true
- Added registry as a module to the opensaber project pom. So, we can issue a single “mvn clean install” directly from the /java folder
- Changes were made earlier to SchemaConfigurator and implemented for RDF. Some remnants are now cleaned up. Also, changed to work with interfaces than specific objects.
- Codacy - (i) Removed un-used vars in the files touched by this PR (ii) some errors like not setting encoding are corrected
- Maven pom seems to be indiscriminately defining and overriding properties - cleansed some that hit my eyes.

**Testing**
- Postman requests with json and ld payloads work - indicating no breakage because of these changes.

**Todo**
- Will test and propose changes for json specific schema validation separately.


java/validators/json/jsonschema/src/main/java/io/opensaber/validators/json/jsonschema/JsonValidationServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+import io.opensaber.validators.IValidate;
+
+public class JsonValidationServiceImpl implements IValidate {
+	private Schema schema;
+
+	private void entity() {
+		InputStream schemaStream = JsonValidationServiceImpl.class.getResourceAsStream("public/_schemas/Teacher.json");
+		JSONObject rawSchema = new JSONObject(new JSONTokener(schemaStream));
+		SchemaLoader schemaLoader = SchemaLoader.builder().schemaJson(rawSchema).draftV7Support()
+				.resolutionScope("http://localhost:8080/_schemas/").build();
```
pritha-tarento: http://localhost:8080 is added in application.yml. Cannot be used as a constant?

java/schema-configuration/src/main/java/io/opensaber/registry/schema/configurator/SchemaConfiguratorFactory.java
```diff
@@ -1,27 +1,43 @@
...
-		} else if (type == SchemaType.SHEX) {
-			schemaConfigurator = shexSchemaConfigurator;
+		try {
+			String schemaFile = environment.getProperty(Constants.FIELD_CONFIG_SCEHEMA_FILE);
+			if (schemaFile == null) {
+				throw new IOException(Constants.SCHEMA_CONFIGURATION_MISSING);
+			}
+			switch (schemaType) {
+			case JSON:
+				schemaConfigurator = new JsonSchemaConfigurator(schemaFile);
```
pritha-tarento: We generally do not use "new" operator in Spring, the beans initiated by configuration class.
So not sure why use of new operator.

java/registry/src/main/java/io/opensaber/registry/config/GenericConfiguration.java
```diff
@@ -177,17 +196,25 @@ public SchemaLoader schemaLoader() throws CustomException, IOException {
...
+		} else {
+		    logger.error("Fatal - not a known validator mentioned in the application configuration.");
+        }
+		return validator;
 	}
 
 	@Bean
 	public SchemaConfiguratorFactory schemaConfiguratorFactory() {
-		return new SchemaConfiguratorFactory();
+		return new SchemaConfiguratorFactory(environment);
```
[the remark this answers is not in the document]
indrajra: On the lines we discussed - The need for this class has been removed. The GenericConfiguration will consume the environment and instantiate the required configurator.

java/validators/json/jsonschema/src/main/java/io/opensaber/validators/json/jsonschema/JsonValidationServiceImpl.java
indrajra: This requires the validator to have a domain set/get method. For now, I've a ticket OS-114 created so that we dont lose track of it.

java/schema-configuration/src/main/java/io/opensaber/registry/schema/configurator/SchemaConfiguratorFactory.java
indrajra: Yes, good catch. Without new, I must have had them as members declared with only one of them being valid and active. That was the thought, anyways this is now gotten rid of.

PR #203, opened 2018-11-13
OS-117 fix for signature entity node creation

Hari-stackroute:
unable to create or update signature for whole entity.
Due to latest code validation changes, it effected existence flow. now it is fixed.

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
+++ b/java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
@@ -119,40 +119,27 @@ public class RegistryServiceImpl implements RegistryService {
 		try {
 			Model signedRdfModel = null;
 			RegistrySignature rs = new RegistrySignature();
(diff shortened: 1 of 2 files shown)


java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -115,44 +115,31 @@ public List getEntityList() {
...
-								entitySignMap, ModelFactory.createDefaultModel());
-						return addEntity(signedRdfModel, subject, property);
-
-					}
-				} else {
-					// else part for json validation
-				}
+			String rootLabel = null;
+
+			// Validating Sign Mandatory data
```
indrajra: can you check persistenceEnabled flag, please?

Hari-stackroute: It is enabled in another addEntity method, which is been called this method.

PR #209, opened 2018-11-22
OS#81 Refactor Request Message and processing across interceptors

indrajra:
At least in our codebase, the interceptors need the body (stream) of the HTTP request, which means the downstream interceptors cannot access it (because, HTTP InputStream cannot be re-read). To overcome this, we have been using attribute map of the HTTP request but still have a problem on ordering of the interceptors. Any new interceptor getting added by an adopter may pick up the body, but fail to populate the right custom attribute name in the map and downstream interceptors will fail.
To prevent all these, it is proposed to have an APIMessage that has a String body, Map placeholder and RequestWrapper objects.
@Autowire APIMessage in every interceptor - (todo: until we create an abstract class that has these)
Todo: There is also a possibility to get rid of Gson in every interceptor.

The changes are along the following lines:
- Every interceptor now as an APIMessage autowired
- The interceptor was handling the message and thus filtering it, which is not expected in an interceptor. It is ok to pass it up.
- Currently the error message is just “ding! Some problem occurred”. Proposed to put the plain message. 

Am sorry that this PR is quite long, but require reviewers to understand the direction


java/pojos/src/main/java/io/opensaber/pojos/APIMessage.java
```diff
@@ -0,0 +1,90 @@
...
+	public Request getRequest() {
+		return request;
+	}
+
+	/**
+	 * Add some temporary request-specific data, say massaged data
+	 * @param key
+	 * @param data
+	 */
+	public void addLocal(String key, Object data) {
```
Hari-stackroute: Instead of addLocal and getLocal, can we rename as getValueFromRequestMap and setValuetoRequestMap

PR #210, opened 2018-11-23
Os 118: Transformation Unifying request and response. 

pritha-tarento:
Changes Done:
Introduced Configuration to support JSON2LD, LD2JSON, LD2LD. This is to avoid conflicts in the design.
ITransformer<T> interface  now comes with 3 implementations respective to the configurations(JSON2LD, LD2JSON, LD2LD).  Hence, transformation is unified to the possible conversions on current scopes and also holds good for future implementations for any other objects. 

Testing Done:
By postman- add API for json/ld both.
By postman- get API for json/ld both.


java/registry-interceptor/src/main/java/io/opensaber/registry/interceptor/RDFConversionInterceptor.java
```diff
@@ -64,4 +67,13 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
 
 		return result;
 	}
+	
+	private Configuration getConfiguration(String contentType){
+		if(contentType.equalsIgnoreCase("application/json")){
```
indrajra: There are good many Http Media type classes to be used here, instead of hardcoded "application/json" string.

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
```diff
@@ -305,4 +322,13 @@
 		return keyToPurge;
 
 	}
+	
+	private Configuration getConfiguration(String accept){
```
indrajra: This code block is repeated. How can this be avoided?

indrajra:
@pritha-tarento - Next item for you to look at is to add LD2RDF as a configuration. We would like to remove RDFConversion as an spring interceptor, but as a normal class invoked based on the media type. This can be invoked from the add entity implementation.

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
pritha-tarento: Done

java/registry-interceptor/src/main/java/io/opensaber/registry/interceptor/RDFConversionInterceptor.java
pritha-tarento: Done

java/middleware-commons/src/main/java/io/opensaber/registry/middleware/util/Constants.java
```diff
@@ -114,6 +114,10 @@ public String getName() {
 		properties, userId, operationType, label, requestId, nodeId, removedRelations, addedRelations, ets, createdAt, transactionData, CREATE, UPDATE, DELETE
 	}
 
+	public enum Direction {	
+		in, out
```
indrajra: Enums are generally best expressed as in upper case - See https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

java/middleware-commons/src/main/java/io/opensaber/registry/transform/TransformerFactory.java
```diff
@@ -1,11 +1,13 @@
 package io.opensaber.registry.transform;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.MediaType;
+
+import io.opensaber.registry.middleware.util.Constants.Direction;
 
-//@Component
 public class TransformerFactory {
```
indrajra: Remove "Factory", simple to call it as Transformer.

java/middleware-commons/src/main/java/io/opensaber/registry/transform/TransformerFactory.java
```diff
@@ -33,29 +34,17 @@
...
-			break;
-
-		default:
-			throw new TransformationException(EXCEPTION_MESSAGE, ErrorCode.UNSUPPOTERTED_TYPE);
-
-		}
-		return transformer;
-	}*/
+	
+	public Configuration getConfiguration(String mediaType, Direction direction){
```
indrajra: I would think Configuration can be built elsewhere, not in this class - Transformer. Can you move this to MediaTypeConfiguration.java or so? The following line looks odd to me. 
`Configuration config = transformerFactory.getConfiguration(contentType, Direction.in);`
From the Transformer, we are getting the configuration and actually passing it back to it to get an appropriate transformer impl.

java/middleware-commons/src/main/java/io/opensaber/registry/transform/TransformerFactory.java
pritha-tarento: done

java/middleware-commons/src/main/java/io/opensaber/registry/middleware/util/Constants.java
pritha-tarento: done

java/middleware-commons/src/main/java/io/opensaber/registry/transform/TransformerFactory.java
indrajra: Sure, I'll go ahead with the merge on this one and look forward to your changes next. Thanks @pritha-tarento .

PR #219, opened 2018-12-07
OS-137 Make Graph transient as a http request

indrajra:
We learnt that we were using Graph throughout the application life scope contrary to the recommendation by the library developers - See https://github.com/SteelBridgeLabs/neo4j-gremlin-bolt/issues/78 for more details.
To abide by the recommendation, the changes here are proposed.

The driver is initialised once and the graph objects working with it are done per request. 

**Other changes**
- Changed log lines to appropriate levels.
- Introduced env var to sharding support

**Testing**:
- A simple add operation works, which means the connection could be caught to the underlying database.
- When I tested with ab, apib the application hung up - due to the way we are using traversals in the current RDF based persistence models. It is known that this hung up will not happen in the proposed simpler data modelling.

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/sink/Neo4jGraphProvider.java
+++ b/java/registry/src/main/java/io/opensaber/registry/sink/Neo4jGraphProvider.java
@@ -29,7 +29,7 @@ public class Neo4jGraphProvider extends DatabaseProvider {
 
 	public Neo4jGraphProvider(DBConnectionInfo connection) {
 		connectionInfo = connection;
-		profilerEnabled = connection.isEnableProfiler();
+		profilerEnabled = connection.isProfilerEnabled();
 		// TODO: Check with auth
 		driver = GraphDatabase.driver(connection.getUri(),
 				AuthTokens.none());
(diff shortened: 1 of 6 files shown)


java/registry/src/main/java/io/opensaber/registry/sink/Neo4jGraphProvider.java
```diff
@@ -62,10 +84,9 @@ public void shutdown() throws Exception {
 		logger.info("**************************************************************************");
 		logger.info("Gracefully shutting down Neo4J GraphDB instance ...");
 		logger.info("**************************************************************************");
-		graph.close();
+
```
Hari-stackroute: I hope closing of graph should be taken care at the end of each transaction right.

indrajra: Yes, it is the responsibility of the callers to close the graph. Since this class doesn't know the list of graph instances created, it cannot do the close itself. 
Am taking liberty to merge this seeing your approval notes.

PR #222, opened 2018-12-11
OS-131: Introduce database shard advisor based on custom property

pritha-tarento:
Indroduced Shard advisor.
Added interface IShardAdvisory. Single sample impl for serial number property.
ShardAdvisor for registering the impl with property, this suppose to get pre loaded.
Shard factory intiatiate a connection with help of ShardAdvisor.
Introduced ShardManager to activate a Shard instance


java/registry/src/main/java/io/opensaber/registry/shard/advisory/SerialNumberShardAdvisor.java
```diff
@@ -0,0 +1,41 @@
...
+			switch (serNo % 2) {
+			case 0:
+				connectionInfo = dBConnectionInfoMgr.getConnectionInfo().get(1);
+				break;
+			case 1:
+				connectionInfo = dBConnectionInfoMgr.getConnectionInfo().get(0);
+				break;
+			}
+		}
+		shardId = connectionInfo.getShardId();
```
indrajra: dont see use of this line.
Also, if possible, simple way to write this code would be `connectionInfo = dbConnectionInfoMgr.getConnectionInfo().get(serNo % 2);`
Or at least rather than using index number of an array which is not referred in the application.yml, can we use the shardId?

java/registry/src/main/resources/application.yml.sample
```diff
@@ -36,17 +36,19 @@ database:
   #provider available are NEO4J, SQLG, TINKERGRAPH, CASSANDRA, ORIENTDB.
   #Note: Please avoid repeating the shardId.
   provider: ${database_provider:NEO4J}
+  shardPrefix: false
+  shardProperty: serialNum
```
indrajra: Extend this to be set as env var

java/registry/src/test/java/io/opensaber/registry/service/impl/RegistryServiceImplTest.java
```diff
@@ -164,7 +182,7 @@ public void test_health_check_up_scenario() throws Exception {
 	public void test_health_check_down_scenario() throws Exception {
 		when(encryptionService.isEncryptionServiceUp()).thenReturn(true);
 		when(mockDatabaseProvider.isDatabaseServiceUp()).thenReturn(false);
-		when(signatureService.isServiceUp()).thenReturn(true);
+		when(signatureService.isServiceUp()).thenReturn(false);
```
indrajra: looks there is no change for signatureService necessary, if you were able to mock the databaseProvider to return false. The health flag is an AND of encryption, signature and database being UP.

pritha-tarento: when(signatureService.isServiceUp()).thenReturn(false) - does not matter,
 I see depends on when(mockDatabaseProvider.isDatabaseServiceUp()).thenReturn(false);

java/registry/src/main/java/io/opensaber/registry/shard/advisory/SerialNumberShardAdvisor.java
pritha-tarento: Choose by shardId.Done.

java/registry/src/main/resources/application.yml.sample
pritha-tarento: Done

java/registry/src/test/java/io/opensaber/registry/service/impl/RegistryServiceImplTest.java
pritha-tarento: Done.

PR #224, opened 2018-12-11
OS-142 - extending array support for node properties and entities

Hari-stackroute:
Persisting the parent node while application loading
Extending the array support for node properties and entities


indrajra:
Approved.

PR #228, opened 2018-12-13
OS-143 Implement Neo4jId creation

indrajra:
Introduced a ‘none’ property that could switch off shards and use a default shard. There is some more scope to ease this. Presently the READ methods in the controller explicitly seek the default shard provider. These methods would be changed - right now I don’t have a way in mind, but sooner will.
Introduced a startTransaction and commitTransaction method - This avoids transaction unsupported failures that happens using stores like TinkerGraph.
Introduced a uuidPropertyName in the config, so that people could change from default ‘osid’ to something else if they want.
Even though the validation was printing in the console about non-compliance of the schema, the validation was always passed. This must not be allowed. Returned a flag based on the occurrence of the validation exceptions.
TPGrapMain - we must change the name sooner, but for now made all methods as non-static. The uuidPropertyName is getting set to null during tests, causing failures. For now masked it with some ‘testuuid’ value assignment.
Returned Vertex or Id on some places to re-use.
Introduced a generic readGraph2Json() that can cut across all database. I believe because we now work with uuids, the reads must be faster. Needs perf testing. Till then the other neo4j specific implementation could also be retained.
And lastly, the provider’s driver object need to be initialised only once per shard. Introduced a local map to preserve that.

**Testing**
- Add and readg (temporal read) API works
- Restarting app doesn't create the group nodes again (which indicates that the uuids are properly fetched).

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/util/TPGraphMain.java
+++ b/java/registry/src/main/java/io/opensaber/registry/util/TPGraphMain.java
@@ -85,16 +85,14 @@ public class TPGraphMain {
         return v1.addEdge(label, v2);
     }
 
-    public static Vertex createParentVertex(Graph graph) {
-        String personsStr = "Persons";
-        String personsId = "ParentEntity_Persons";
+    public static Vertex createParentVertex(Graph graph, String parentLabel) {
         GraphTraversalSource gtRootTraversal = graph.traversal();
-        GraphTraversal<Vertex, Vertex> rootVertex = gtRootTraversal.V().hasLabel(personsStr);
+        GraphTraversal<Vertex, Vertex> rootVertex = gtRootTraversal.V().hasLabel(parentLabel);
         Vertex parentVertex = null;
         if (!rootVertex.hasNext()) {
-            parentVertex = graph.addVertex(personsStr);
-            parentVertex.property("id", personsId);
-            parentVertex.property("label", personsStr);
+            parentVertex = graph.addVertex(parentLabel);
+            // TODO: this could be parentVertex.id() after we make our own Neo4jIdProvider
+            parentVertex.property("osid", parentLabel + "osid");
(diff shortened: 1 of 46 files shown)


java/registry/src/test/java/io/opensaber/registry/dao/impl/RegistryDaoImplTest.java
```diff
@@ -23,6 +23,8 @@
 import java.util.NoSuchElementException;
 import java.util.UUID;
 
+import io.opensaber.registry.model.DBConnectionInfo;
```
Hari-stackroute: seems unused import

java/registry/src/main/java/io/opensaber/registry/util/TPGraphMain.java
```diff
@@ -19,149 +20,145 @@
...
-                parentVertex = graph.addVertex(parentLabel);
-                // TODO: this could be parentVertex.id() after we make our own Neo4jIdProvider
-                parentVertex.property("osid", UUID.randomUUID().toString());
-                tx.commit();
-            } else {
-                parentVertex = graph.addVertex(parentLabel);
-                // TODO: this could be parentVertex.id() after we make our own Neo4jIdProvider
-                parentVertex.property("osid", UUID.randomUUID().toString());
-            }
+            parentVertex = graph.addVertex(parentLabel);
```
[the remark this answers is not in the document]
indrajra: The transaction management is done outside this class.

java/registry/src/test/java/io/opensaber/registry/dao/impl/RegistryDaoImplTest.java
indrajra: will do next. I want you guys to have this code before I make further changes.

PR #230, opened 2018-12-14
OS-143 Improve write performance by fetching parentVertex upfront

indrajra:
EntityParenter now has ensure methods for parent vertices and has new methods for fetching the parentVertex.
It is proven by our PoC that Vertex object can be re-used across multiple Graph objects and it works.
Introduced label generators to make code separate and cleaner.
Set root level of log to debug for us to pay close attention.

With debug application and debug level of logging, writes are at 36 RPS. Reads are not that great, but we know we can improve by warming up a cache.


indrajra:
There are 9 new issues reported due to un-used params. These will be cleansed in another ticket that addresses the refactoring part; sorry about that.

PR #234, opened 2018-12-18
Os 151 Read and Add Api:  EntityCache enabled

pritha-tarento:
EntityCache is now being used by registry controller.
Changes Read Api:
EnitityCache: getShard(recordId) to get the shard id for existing record in database.
Changes Add Api:
ShardDatabaseProvider Pojo is introduced for holding the shard-databaseProvider map. Modification is done accordingly ShardManager:getDatabaseProvider() to get ShardDatabaseProvider object. Benefit: it holds the one to one relation for shard identifier and the databaseProvider.
EntityCache is facilated to add a shard-record map as after creating and persisting the record in database.
Added test cases for EntityCache
Resolved issue commit Transaction. Autocloseable for Neo4jDatabaseProvider
TODO:
To fix broken test cases with removing setters for databaseProvider for all layers.


java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
```diff
@@ -259,17 +263,22 @@
 		String jsonString = apiMessage.getRequest().getRequestMapAsString();
 		String entityType = apiMessage.getRequest().getEntityType();
 
+		ShardDatabaseProvider shardDbProvider = null;
 		try {
 			if (shardManager.getShardProperty().compareToIgnoreCase(Constants.NONE_STR) != 0) {
 				int slNum = (int) ((HashMap<String, Object>) apiMessage.getRequest().getRequestMap().get(entityType))
```
indrajra: These statements are long. Can you declare vars and make it short.

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
```diff
@@ -259,17 +263,22 @@
...
 						.get(shardManager.getShardProperty());
-				databaseProviderWrapper.setDatabaseProvider(shardManager.getDatabaseProvider(slNum));
+				shardDbProvider = shardManager.getDatabaseProvider(slNum);
```
indrajra: I see we may be repeating this if-else block. Can we wrap this into a class?

java/registry/src/main/java/io/opensaber/registry/sink/shard/ShardManager.java
```diff
@@ -3,52 +3,70 @@
...
+		logger.info("Activate db shard "+connectionInfo.getShardId()+" for attribute value "+attributeValue);
+		return shardDbProvider;
 	}
 
 	public String getShardProperty() {
 		return dbConnectionInfoMgr.getShardProperty();
 	}
 
-	public DatabaseProvider getDatabaseProvider(Object attributeValue) throws CustomException {
+	public ShardDatabaseProvider getDatabaseProvider(Object attributeValue) throws CustomException {
```
indrajra: This is confusing. What's the difference b/w ShardDatabaseProvider and DatabaseProvider?

java/registry/src/main/java/io/opensaber/registry/util/EntityCache.java
```diff
@@ -44,5 +45,27 @@ public String getShard(String recordId) throws IOException {
 		}
 		return shardId;
 	}
+	/**
+	 * Cache to add with a new shard record mapping
+	 * @param shardId
+	 * @param recordId
+	 * @return
+	 */
+	public boolean addShardRecord(String shardId, String recordId){
```
indrajra: Name it simply as this addEntity() and currently you're passing just the recordId. Can you wrap this in a class and call it Entity. This class will have recordId and will also likely contain the complete Json object (what was read)...What's the fun in using the cache only to figure out the shard? Lets not hit the database too.

java/registry/src/test/java/io/opensaber/registry/util/EntityCacheTest.java
```diff
@@ -0,0 +1,71 @@
...
+		entityCache.getShard("UUID_NOT_PRESENT");
+
+	}
+
+	@Test
+	public void testGetShardforPresentRecord() throws IOException {
+		assertEquals("Shard1", entityCache.getShard("UUID1"));
+		assertEquals("Shard1", entityCache.getShard("UUID2"));
+
+	}
```
indrajra: Should we have another test and say for "Shard1" we have the following UUIDs. Both ways I mean.

java/registry/src/main/java/io/opensaber/registry/util/EntityCache.java
pritha-tarento: Done - as renaming. Adding cache for Object will be taken care of as a part of other PR

java/registry/src/main/java/io/opensaber/registry/sink/shard/ShardManager.java
pritha-tarento: renamed ShardDatabaseProvider as Shard

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
pritha-tarento: Done. Shard manager in exposed for other layers.

java/registry/src/test/java/io/opensaber/registry/util/EntityCacheTest.java
pritha-tarento: Done.

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
pritha-tarento: Done

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
```diff
@@ -263,22 +262,22 @@
...
-				databaseProviderWrapper.setDatabaseProvider(shardDbProvider.getDatabaseProvider());
-			} else {
-				shardDbProvider = shardManager.getDefaultDatabaseProvider();
-				databaseProviderWrapper.setDatabaseProvider(shardDbProvider.getDatabaseProvider());
+			Map requestMap = ((HashMap<String, Object>) apiMessage.getRequest().getRequestMap().get(entityType));
+			logger.info("Add api: entity type " + requestMap + " and shard propery: " + shardManager.getShardProperty());
+			
+			logger.info("request: "+requestMap.get(shardManager.getShardProperty()));
+			Object attribute = null;
+			if(requestMap.containsKey(shardManager.getShardProperty())){
```
indrajra: Start using requestMap.getOrDefault() - avoid if.

java/registry/src/main/java/io/opensaber/registry/dao/TPGraphMain.java
```diff
@@ -153,18 +153,18 @@ public String processEntity(Graph graph, JsonNode node) {
...
-        GraphTraversal<Vertex, Vertex> graphTraversal = graph.traversal().V().hasLabel(predicateStr);
-        while (graphTraversal.hasNext()){
-            Vertex v = graphTraversal.next();
-            if (v != null) {
-                uuids.add(v.value(uuidPropertyName).toString());
-            }
-        }
-        return uuids;
-    }
+	public List<String> getUUIDs(Graph graph, Set<String> labels) {
```
indrajra: I think this will be taken from 2.0.0 and needs no change.

java/registry/src/main/java/io/opensaber/registry/sink/shard/Shard.java
```diff
@@ -0,0 +1,36 @@
...
+	
+	public void setShardId(String shardId) {
+		this.shardId = shardId;
+	}
+
+	public void setDatabaseProvider(DatabaseProvider databaseProvider) {
+		this.databaseProvider = databaseProvider;
+	}
+
+	/*	public Shard(String shardId, DatabaseProvider databaseProvider){
```
indrajra: Get rid of this this.

java/registry/src/main/java/io/opensaber/registry/sink/shard/ShardManager.java
```diff
@@ -24,49 +24,73 @@
...
 	}
 	/**
-	 * Get a databaseProvider given a shardId from entity cache
+	 * activate a shard given a shardId from entity cache
 	 * @param shardId
 	 * @return
 	 */
-	public DatabaseProvider getDatabaseProvider(String shardId) throws CustomException{
-		DatabaseProvider databaseProvider = null;
+	public void activateShard(String shardId) throws CustomException{
```
indrajra: This was private earlier. Why did this change? What if the callers forget to activate, but simply call getShard()?

java/registry/src/test/java/io/opensaber/registry/util/EntityCacheTest.java
```diff
@@ -62,10 +67,23 @@ public void testGetShardForRecordValueNotPresent() throws IOException {
...
-
+	}
+	
+	@Test
+	public void testGetShardTwoforPresentRecord() throws IOException {
+		assertEquals("Shard2", entityCache.getShard("UUID3"));
+		assertEquals("Shard2", entityCache.getShard("UUID4"));
+	}
+	@Test
+	public void testAddEntityForExistingShard(){
```
indrajra: What's the use of these tests? I think these tests will never fail and so not useful to maintain. Why are there 2 asserts in the previous test cases? There must be 1 and only 1. There could be 2 only if there are two things that define the pass criteria.

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
pritha-tarento: Done

java/registry/src/main/java/io/opensaber/registry/sink/shard/Shard.java
pritha-tarento: Done

java/registry/src/test/java/io/opensaber/registry/util/EntityCacheTest.java
pritha-tarento: okay. Done

java/registry/src/main/java/io/opensaber/registry/sink/shard/ShardManager.java
pritha-tarento: Removed getShard(), Not using it anywhere. Done.

java/registry/src/main/java/io/opensaber/registry/dao/TPGraphMain.java
pritha-tarento: This method P.within is not returning all expected items whereas it return only one record.
It was not working as expected, so loading all the uuids.

indrajra: am afraid.

PR #235, opened 2018-12-19
OS 147 - Updating the entity

Hari-stackroute:
supports for updating the parent entity
supports for updating the child entity
Test case for update entity

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
+++ b/java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
@@ -113,7 +113,7 @@ public class RegistryController {
     }
 
     @ResponseBody
-    @RequestMapping(value = "/update", method = RequestMethod.POST)
+    @RequestMapping(value = "/update2", method = RequestMethod.POST)
     public ResponseEntity<Response> update() {
         Model rdf = (Model) apiMessage.getLocalMap(Constants.RDF_OBJECT);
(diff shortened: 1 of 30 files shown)


java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
```diff
@@ -318,6 +318,30 @@
...
 	}
 
+    @ResponseBody
+    @RequestMapping(value = "/update", method = RequestMethod.POST)
+    public ResponseEntity<Response> updateTP2Graph() {
+        ResponseParams responseParams = new ResponseParams();
+        Response response = new Response(Response.API_ID.UPDATE, "OK", responseParams);
+        Map<String, Object> result = new HashMap<>();
+        String jsonString = apiMessage.getRequest().getRequestMapAsString();
+        String entityType = apiMessage.getRequest().getEntityType();
```
indrajra: This var seems unused.

java/registry/src/main/java/io/opensaber/registry/dao/TPGraphMain.java
```diff
@@ -40,6 +43,9 @@
     @Autowired
     private ISchemaConfigurator schemaConfigurator;
 
+    @Autowired
```
indrajra: Why is this autowired here? shouldn't this be a concern of the serviceImpl class?

PR #242, opened 2018-12-21
OS#146 Remove RDF, SHEX related files and code

indrajra:
Most of these are file deletions - .jsonld and .shex
The test cases is another area we need some focussed efforts that we will gain grounds in the coming weeks.
The /add and /read works after doing all the above.


indrajra:
Since there is merely removal of code, going without explicit approval.

PR #244, opened 2018-12-24
Issue OS-138 code change for sign api which is used for signing entit…

Hari-stackroute:
Code change for sign api which is used for signing entity and value which is provided as input data
Example input:
{
  "id":"open-saber.utils.sign",
  "ver":"1.0",
  "ets":"11234",
  "params":{
    "did":"",
    "key":"",
    "msgid":""
  },
  "request":{
	"value":["sdfsfsfs","dsdss"]
   }
}

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/controller/RegistryUtilsController.java
+++ b/java/registry/src/main/java/io/opensaber/registry/controller/RegistryUtilsController.java
@@ -62,10 +62,9 @@ public class RegistryUtilsController {
 		try {
 			watch.start("RegistryUtilsController.generateSignature");
 			Map<String, Object> requestBodyMap = apiMessage.getRequest().getRequestMap();
-			if (requestBodyMap.containsKey(Constants.REQUEST_ATTRIBUTE)
-					&& requestBodyMap.containsKey(Constants.ATTRIBUTE_NAME)) {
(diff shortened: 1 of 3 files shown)


java/registry/src/main/java/io/opensaber/registry/controller/RegistryUtilsController.java
```diff
@@ -62,10 +62,9 @@
 		try {
 			watch.start("RegistryUtilsController.generateSignature");
 			Map<String, Object> requestBodyMap = apiMessage.getRequest().getRequestMap();
-			if (requestBodyMap.containsKey(Constants.REQUEST_ATTRIBUTE)
-					&& requestBodyMap.containsKey(Constants.ATTRIBUTE_NAME)) {
+			if (null !=requestBodyMap && (requestBodyMap.containsKey(Constants.SIGN_ENTITY) || requestBodyMap.containsKey(Constants.SIGN_VALUE))){
 				Object result = signatureService
-						.sign(gson.fromJson(requestBodyMap.get(Constants.ATTRIBUTE_NAME).toString(), mapType));
+						.sign(gson.fromJson(requestBodyMap.toString(), mapType));
```
indrajra: why can't the .sign() take requestBodyMap if it requires a mapType.
Also, isn't there any changes for /verify and /encrypt endpoints - please ensure all /utils/<APIs> are tested.

Hari-stackroute: I will test the sign with as you suggested. I didn't seen the remaining api's, i will check and fix if required.

PR #248, opened 2019-01-02
OS 150 - Signing the data

Hari-stackroute:
Signing the whole entity for add api
Signing the whole entity in update api.


java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
```diff
@@ -265,16 +266,13 @@
...
-
-        String dataObject = apiMessage.getRequest().getRequestMapAsString();
-        String entityType = apiMessage.getRequest().getEntityType();
-        JsonNode reqJsonNode = apiMessage.getRequest().getRequestMapNode();
-        String label = apiMessage.getRequest().getRequestMap().get(dbConnectionInfoMgr.getUuidPropertyName()).toString();
-
-        RecordIdentifier recordId = RecordIdentifier.parse(label);
-        String shardId = dbConnectionInfoMgr.getShardId(recordId.getShardLabel());
 
+		String dataObject = apiMessage.getRequest().getRequestMapAsString();
```
indrajra: looks you're using dataObject. If so, remove.

java/registry/src/main/java/io/opensaber/registry/service/SignatureHelper.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Map;
+
+@Component
+public class SignatureHelper {
+    @Autowired
+    private SignatureService signatureService;
+    @Autowired
+    private ObjectMapper objectMapper;
+    @Value("${registry.rootEntity.type}")
+    private String registryRootEntityType;
```
indrajra: Please don't use this property. This will have to be removed. We introduced for release 1.0.0 on time - as a workaround only.

java/registry/src/main/java/io/opensaber/registry/service/SignatureHelper.java
```diff
@@ -0,0 +1,58 @@
...
+    @Value("${signature.keysURL}")
+    private String signatureKeyURl;
+
+    /** Signs the entity and returns the entity with signed json appending
+     * @param rootNode
+     * @return
+     * @throws SignatureException.UnreachableException
+     * @throws SignatureException.CreationException
+     */
+    public JsonNode getSignedJson(JsonNode rootNode) throws SignatureException.UnreachableException, SignatureException.CreationException {
```
indrajra: please create smaller fine methods - sign(), verify(), addSignature() - this is current getSignedJson what you have, createSignatureItem() - this is your current mergeSignData().

java/registry/src/main/java/io/opensaber/registry/service/SignatureHelper.java
```diff
@@ -0,0 +1,58 @@
...
+        Map<String, Object> signMap = (Map<String, Object>) signatureService.sign(signReq);
+        mergeSignData(rootNode, signMap);
+        return signedRoot;
+    }
+
+    /** Merges sign data to entity json
+     * @param rootNode
+     * @param signMap
+     */
+    private void mergeSignData(JsonNode rootNode, Map<String, Object> signMap) {
```
indrajra: In addition to earlier, pass the rootNode type as string...this method doesn't need to have the entire node object.

java/registry/src/main/java/io/opensaber/registry/controller/RegistryController.java
Hari-stackroute: This is the input for service layer and renamed to jsonString, which contains the input for update.

java/registry/src/main/java/io/opensaber/registry/service/SignatureHelper.java
Hari-stackroute: As of that for checking the root label of entity, most of the places we have used this property for checking. So this can't be removed as of now. we will take it later.

java/registry/src/main/java/io/opensaber/registry/service/SignatureHelper.java
Hari-stackroute: Ok. renamed to signJson and mergeSign.

java/registry/src/main/java/io/opensaber/registry/service/SignatureHelper.java
Hari-stackroute: Now sending proper json node instead of whole json and merging the sign data.

java/registry/src/main/java/io/opensaber/registry/service/SignatureHelper.java
indrajra: We must get rid of this. I sense from our discussion that this requires some more efforts and so okay to approve.

java/registry/src/main/java/io/opensaber/registry/service/SignatureHelper.java
indrajra: sign and mergeSign would have been much meaningful...

PR #249, opened 2019-01-02
Os 157  Remove schema configurator

pritha-tarento:
Introduced SchemaDefination and DefinationManager which holds the all the definitions.
Application available schema defination are loading while boot which a default behaviour.
Remove the schema-configurator module with related dependencies to be fixed. 
TODO:
#Currently to call the SchemaDefination:getPrivateFields() used hardcoded value "teacher".
#Testcases


java/registry/src/main/java/io/opensaber/registry/util/DefinationNode.java
```diff
@@ -0,0 +1,8 @@
+package io.opensaber.registry.util;
+
+public class DefinationNode {
```
indrajra: Spell check....why the name Node here? Please write javadoc at class level to start with - let me understand what you're proposing.

java/registry/src/main/java/io/opensaber/registry/util/DefinitionsManager.java
```diff
@@ -1,39 +1,69 @@
...
+            ByteArrayOutputStream result = new ByteArrayOutputStream();
+            byte[] buffer = new byte[1024];
+            int length;
+            while ((length = in.read(buffer)) != -1) {
+                result.write(buffer, 0, length);
             }
+            return result.toString(StandardCharsets.UTF_8.name());
+
+        } catch (IOException e) {
+            logger.error("Cannot load resource " + resource);
```
indrajra: will this print the resource name?

java/registry/src/main/java/io/opensaber/registry/util/SchemaDefination.java
```diff
@@ -0,0 +1,67 @@
...
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SchemaDefination {
+    private static Logger logger = LoggerFactory.getLogger(SchemaDefination.class);
+    private final static String TITLE ="title";
+    private final static String OSCONFIG ="_osconfig";
```
indrajra: Since _osconfig is an internal POJO, can we codify it so? It contains two JsonNode (more specifically ArrayNodes), namely privateFields and signedFields.
Since this is user-related, should we name it as privateAttributes or privateProperties? I mean using attributes, properties instead of fields.

java/registry/src/main/java/io/opensaber/registry/util/DefinationNode.java
pritha-tarento: Removed this file

java/registry/src/main/java/io/opensaber/registry/util/DefinitionsManager.java
pritha-tarento: Done

java/registry/src/main/java/io/opensaber/registry/util/SchemaDefination.java
pritha-tarento: osconfig is an internal POJO is done. I feel we can have like privateFields understandable.

pritha-tarento:
The commits went inconsistent (messed up) and  will open this as a separate PR for ease review.

PR #250, opened 2019-01-02
Os 158 Validation for shard configurations

pritha-tarento:
Introduced ValidConnectionInfo annotation class. Also introduced implementation class shardValidator which implements <ConstraintValidatorValidConnectionInfo, Object>
DBConnectionInfo class is now annotated by custom ValidConnectionInfo annotation. 
The implements empty check for shardId, shardLabel, uri property.
Validation functionality works as:
DatabseConnectionInfoMgr Properties: uuidPropertyName, provider empty not allowed  & ConnectionInfo list size not less than one.
DatabseConnectionInfo properties: shardId  empty not allowed & shardLabel empty not allowed only when connections size > 1


java/registry/src/main/java/io/opensaber/registry/sink/shard/ValidConnectionInfo.java
```diff
@@ -0,0 +1,28 @@
...
+import javax.validation.Constraint;
+import javax.validation.Payload;
+
+@Documented
+@Target({ METHOD, FIELD, CONSTRUCTOR, PARAMETER, TYPE_USE })
+@Retention(RUNTIME)
+@Constraint(validatedBy = { ShardValidator.class })
+public @interface ValidConnectionInfo {
+
+	String message() default "{ConnectionInfo property does not contain value for sharfId or shardLabel or uri}";
```
indrajra: spell check

java/registry/src/main/java/io/opensaber/registry/sink/shard/ShardValidator.java
```diff
@@ -0,0 +1,23 @@
+package io.opensaber.registry.sink.shard;
+
+import io.opensaber.registry.model.DBConnectionInfo;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ShardValidator implements ConstraintValidator<ValidConnectionInfo, Object> {
```
indrajra: Do you want to call this ShardValidator or ConnectionInfoValidator or DBConfigValidator?
I think uuid propertyname, database providers are easy nice extensions that could be added.

pritha-tarento: Renamed to ConnectionInfoValidator .
propertyname, database providers can be added as seperate validator.

java/registry/src/main/java/io/opensaber/registry/sink/shard/ValidConnectionInfo.java
pritha-tarento: Done

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.validation.ConstraintValidatorContext;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ConnectionInfoValidator implements ConstraintValidator<ValidConnectionInfo, DBConnectionInfoMgr> {
+
+	@Override
+	public boolean isValid(DBConnectionInfoMgr mgr, ConstraintValidatorContext context) {
+
+		boolean isValid = true;
```
indrajra: I think you start looking at alternate way of using this boolean. In every if condition, there is a statement assigning the var to false. Can you init with false and then assume if you've hit the end return statement, it could be true? This involves changing conditions, I know.

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
```diff
@@ -0,0 +1,56 @@
...
+		String message = null;
+		if (mgr.getProvider().isEmpty() || mgr.getUuidPropertyName().isEmpty()) {
+			isValid = false;
+			message = "database.provider or database.uuidPropertyName is empty";
+		}
+		if (mgr.getConnectionInfo().size() < 1) {
+			isValid = false;
+			message = "database.connectioninfo[0] not found";
```
indrajra: Add - At least one connectionInfo must be specified.

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
```diff
@@ -0,0 +1,56 @@
...
+		}
+		for (DBConnectionInfo info : mgr.getConnectionInfo()) {
+			if (info.getShardId().isEmpty() || info.getShardLabel().isEmpty() || info.getUri().isEmpty()) {
+				isValid = false;
+				message = "database.connectionInfo.shardId or database.connectionInfo.shardLabel or database.connectionInfo.uri is empty";
+				break;
+			}
+			if (notUnique(mgr.getConnectionInfo(), info.getShardId(), info.getShardLabel())) {
```
indrajra: Don't call a function name in negative...say isUnique instead.

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
```diff
@@ -0,0 +1,56 @@
...
+				isValid = false;
+				message = "database.connectionInfo.shardId and database.connectionInfo.shardLabe must be unique";
```
indrajra: Labe -> Label....say this clear - are you expecting id and label individually to be unique or a combination of them to be unique?

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
```diff
@@ -0,0 +1,56 @@
...
+				break;
+			}
+		}
+
+		if (!isValid)
```
indrajra: I see this condition un-necessary...if message is null/empty, so be it. Why bother when you have the flag? May be in the setMessage you can check for nullness - to avoid any problems with Spring function.

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
```diff
@@ -0,0 +1,56 @@
...
+
+		return isValid;
+	}
+
+	private void setMessage(ConstraintValidatorContext context, String message) {
+		context.disableDefaultConstraintViolation();
+		context.buildConstraintViolationWithTemplate(message).addConstraintViolation();
+	}
+
+	private boolean notUnique(final List<DBConnectionInfo> list, final String shardId, String shardLabel) {
```
indrajra: list is not an appropriate var name...

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.opensaber.registry.model.DBConnectionInfo;
+import io.opensaber.registry.model.DBConnectionInfoMgr;
+import java.util.List;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ConnectionInfoValidator implements ConstraintValidator<ValidConnectionInfo, DBConnectionInfoMgr> {
```
indrajra: Now that this has expanded to check for 'database' section of the config, won't you rename the class so? like, DatabaseConfigValidator.

java/registry/src/test/java/io/opensaber/config/validation/DBConnectionInfoMgrTest.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Set;
+import javax.validation.ConstraintViolation;
+import javax.validation.Validation;
+import javax.validation.Validator;
+import javax.validation.ValidatorFactory;
+import org.junit.Before;
+import org.junit.Test;
+
+public class DBConnectionInfoMgrTest {
+
```
indrajra: Good tests, remove commented lines and add a test for shard labels also (even if they are same as shard ids).

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
pritha-tarento: Done. Shard Id & label uniqueness is checked individually.

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
pritha-tarento: isvalid flag default ->false. checked by message if null set isvalid flag true.
Done

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
pritha-tarento: Ok done

java/registry/src/main/java/io/opensaber/registry/config/validation/ConnectionInfoValidator.java
pritha-tarento: Done

java/registry/src/test/java/io/opensaber/config/validation/DBConnectionInfoMgrTest.java
pritha-tarento: Added a more testcase for shardlabel. Done

PR #258, opened 2019-01-14
OS-167 Fix numbers, boolean to return as is instead of strings

indrajra:
- Created ValueType helper class to assist in reading/writing appropriate value types
- Created a boolean and double type member in Person, Vehicle respectively and ensured both add and read work with these types.
- Prior to this fix, a serialNum of integer type is used to add; in return for read, this will end up string. Now boolean, double, int and string types are restored for both reads and writes.
- Alongside, had to fix the following:
1. fixed hard-coded need for root entity type from the application config. This was fixed in all places in 2.0.0, but was missed from signature helper.
2. the graph session management was broken leading to graph object closures in postgres.


java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -212,84 +210,85 @@ public void updateEntity(String id, String jsonString) throws Exception {
...
-                    if ((inputNodeVertex.property(Constants.STATUS_KEYWORD).isPresent() &&
-                            inputNodeVertex.property(Constants.STATUS_KEYWORD).value().equals(Constants.STATUS_INACTIVE))) {
-                        throw new RecordNotFoundException("Cannot perform the operation");
-                    }
-                    if(inputNodeVertex.property(Constants.ROOT_KEYWORD).isPresent()){
-                        rootVertex = graph.vertices(inputNodeVertex.property(Constants.ROOT_KEYWORD).value()).next();
-                    } else {
-                        rootVertex = inputNodeVertex;
-                    }
+            Transaction tx = databaseProvider.startTransaction(graph);
```
Hari-stackroute: reason behind creating transaction without try block is in commit transaction it is being closed right? what is some exception occurrence in between start and commit statements?

indrajra: I just did this for consistency - startTransaction followed by commitTransaction. 
Otherwise the transaction is autocloseable, but I wanted to route it through explicit commitTransaction. If we just used try, we could argue for not having an explicit commit as well.

PR #259, opened 2019-01-16
OS 166 - Support for updating sub-entity

Hari-stackroute:
Update subentity
changed direction of signature node of entity
filtering nodes while read and update api's


Hari-stackroute:
> Let the signature direction not change. We must allow document level signatures to be passed back. We can filter it using other mechanism in future, but current implementation must pass it back.

Modified as suggested. Direction is set as IN, read, add, update api's modified.

PR #267, opened 2019-01-30
OS#175 Update refactoring

indrajra:
- Cleared up Configuration - 
    - we don’t support LD2LD, but json2json and it was confusing to have multiple live configurations. We can take in only json and spit out either a json or LD and that’s it for now.
    - Accordingly, simplified functions for getting this configuration
- Renamed tp2GraphMain to registryDaoImpl and other methods in the controller
- Fixed the “entity” in response to add to call out the entity real name, like “Teacher”. For example, an add of “Person” record will respond "result”:{“Person”:{“osid":"1-ef1f3f2e-7f1f-4d9f-954e-185ee61e9d4e"}
- Created a ReadConfiguratorFactory to construct read configurations easily in one-liner when needed.
- Whenever id().toString() is used, there is likely a problem in database such as Postgres. We should use getId() and generateId() to be safe. This will also fix the read problem, @Hari encountered on Postgres is 2.0.1
- Added a debug log in logback.xml for us to get debug in console (could be changed to info when releasing)
- While testing, observed some hangs and attribute that to indexing in neo4j. Temporarily disabling that code in this PR.
- We want to get rid of Shard usage in registryDaoImpl - one of the two variants of getEntity() has this change. Need to do for the other. There are several todos left in update that need to be taken up next (almost immediately).

**Testing**
1. The 2 payloads below were fired for update. The records were read back via APIs and ensured the right contents came up.
```
"Teacher": {
    	"osid": "1-d8f74af8-0897-4c84-8715-f0b6a22a019b",
    	"inServiceTeacherTrainingFromBRC": {
                "osid": "1-13538e3d-89dc-4dda-b18e-fbfc18d68446",
                "daysOfInServiceTeacherTraining": "20"
            }
    }

```
```
"Teacher": {
    	"osid": "1-d8f74af8-0897-4c84-8715-f0b6a22a019b",
    	"teacherName": "Rajesh"
    }
```
2. When reading with "includeSignatures=true" observed that the signatures are returned.
3. The above were tested with both neo4j and postgres.


java/registry/src/main/java/io/opensaber/registry/util/ReadConfiguratorFactory.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public static ReadConfigurator getWithSignatures() {
+        ReadConfigurator configurator = new ReadConfigurator();
+        configurator.setIncludeSignatures(false);
+        configurator.setIncludeTypeAttributes(false);
+        return configurator;
+    }
+
+    public static ReadConfigurator getOne(boolean withSignatures) {
```
Hari-stackroute: getDefault and getWithSignatures both does the same, why? if both are public and accessible why getOne?

PR #268, opened 2019-01-30
Os 176 Bug while creating index for databaseProvider 

pritha-tarento:
Issue below:
1.Read API neo4jGraph throwing exception: Connection to the database was lost because someone called `interrupt()
2. Index Field property values were getting duplicating while adding new index field value.

As a part of this PR:
Introduced Indexer which  included 
+ createIndex(), setIndexFields(), setUniqueIndexFields
  updateIndices(), - createNonUniqueIndex(), - createUniqueIndex()

Introduced IndexHelper which included 2 methods
+ isIndexPresent() , getNewFields()

Added 2 methods in EntityParenter:
- asyncAddIndex() 
ensureIndexExists() : checks isIndexPresent()  and calls for asyncAddIndex()

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
+++ b/java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
@@ -201,13 +201,17 @@ public class RegistryServiceImpl implements RegistryService {
      * @param label
      * @throws Exception
      */
-    private void addIndex(DatabaseProvider dbProvider, String shardId, String label) throws Exception{
-        try (OSGraph osGraph = dbProvider.getOSGraph()) {
-            Graph graph = osGraph.getGraphStore();
-            Transaction tx = dbProvider.startTransaction(graph);           
-            ensureIndexExists(graph, dbProvider, shardId, label);
-            dbProvider.commitTransaction(graph, tx);
-        }            
+    private void addIndex(DatabaseProvider dbProvider, String shardId, String label) {
+        new Thread(() -> {
+            try (OSGraph osGraph = dbProvider.getOSGraph()) {
+                Graph graph = osGraph.getGraphStore();
+                Transaction tx = dbProvider.startTransaction(graph);           
+                ensureIndexExists(graph, dbProvider, shardId, label);
+                dbProvider.commitTransaction(graph, tx);
+            } catch (Exception e) {
+              logger.info("Can't create index on table " + label);
+          }
+        }).start();           
     }
 
     /**
--- a/java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
+++ b/java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
@@ -189,22 +189,38 @@ public class RegistryServiceImpl implements RegistryService {
 
                 vertexLabel = rootNode.fieldNames().next();
             }
-            ensureIndexExists(dbProvider, shard.getShardId(), vertexLabel);
+            addIndex(dbProvider, shard.getShardId(), vertexLabel);
         }
 
         return entityId;
+    }
+    /**
+     * Adds a index to the given label(vertex/table)
+     * @param dbProvider
+     * @param shardId
+     * @param label
+     * @throws Exception
+     */
+    private void addIndex(DatabaseProvider dbProvider, String shardId, String label) throws Exception{
+        try (OSGraph osGraph = dbProvider.getOSGraph()) {
+            Graph graph = osGraph.getGraphStore();
+            Transaction tx = dbProvider.startTransaction(graph);           
+            ensureIndexExists(graph, dbProvider, shardId, label);
+            dbProvider.commitTransaction(graph, tx);
+        }            
     }
 
     /**
      * Ensures index for a vertex exists Unique index and non-unique index is
      * supported
      * 
+     * @param graph
      * @param dbProvider
      * @param shardId
      * @param label
      *            a type vertex label (example:Teacher)
      */
-    private void ensureIndexExists(DatabaseProvider dbProvider, String shardId, String label) {
+    private void ensureIndexExists(Graph graph, DatabaseProvider dbProvider, String shardId, String label) {
 
         Vertex parentVertex = entityParenter.getKnownParentVertex(label, shardId);
         Definition definition = definitionsManager.getDefinition(label);
@@ -217,7 +233,7 @@ public class RegistryServiceImpl implements RegistryService {
         Indexer indexer = new Indexer(dbProvider);
         indexer.setIndexFields(indexFields);
         indexer.setUniqueIndexFields(indexUniqueFields);
-        indexer.createIndex(label, parentVertex);
+        indexer.createIndex(graph, label, parentVertex);
 
     }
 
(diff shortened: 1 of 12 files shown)


java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -238,18 +238,35 @@ private void ensureIndexExists(DatabaseProvider dbProvider, Graph graph, String
      * @return
      */
     private boolean indexFieldsExists(Vertex parentVertex, List<String> fields) {
+        boolean contains = false;
+        if (parentVertex.property(Constants.INDEX_FIELDS).isPresent()) {
+            for (String field : fields) {
```
indrajra: Can you get rid of for loop and simply use `contains = fields.contains(parentVertex)` Overall I see not much difference between indexFieldsExists and the fieldExists - both achieve the same thing and there is a lot of code here. Please reduce.

java/registry/src/main/java/io/opensaber/registry/sink/Neo4jGraphProvider.java
```diff
@@ -26,6 +26,8 @@
     private boolean profilerEnabled;
     private DBConnectionInfo connectionInfo;
     private Neo4jIdProvider neo4jIdProvider = new Neo4jIdProvider();
+    private Neo4JGraph neo4jGraph;
```
indrajra: This is disallowed. Neo4JGraph we know cannot be persisted for more than 1 api request. It must be within methods only.

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
pritha-tarento: indexFieldsExists (for list) uses  fieldExists(for a single field), there was one line repetation I removed.
String:contains() method is not efficient so I have used -Arrays.stream(indexFields).anyMatch(field::equals)

pritha-tarento: Done.

java/registry/src/main/java/io/opensaber/registry/sink/Neo4jGraphProvider.java
pritha-tarento: Done

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -201,28 +202,26 @@ public String addEntity(String jsonString) throws Exception {
...
             if (!indexFieldsExists(parentVertex, indexFields)){
-                dbProvider.createIndex(label, indexFields);
                 setPropertyValuesOnParentVertex(parentVertex, indexFields);
+                dbProvider.createIndex(label, indexFields);
 
             }
             if(!indexFieldsExists(parentVertex, indexUniqueFields)){
-                dbProvider.createUniqueIndex(label, indexUniqueFields);
                 setPropertyValuesOnParentVertex(parentVertex, indexUniqueFields);
+                dbProvider.createUniqueIndex(label, indexUniqueFields);
```
indrajra: Why is this moved? What happens if we set the property, but failed to create index for whatever reason?

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -239,13 +238,12 @@ private void ensureIndexExists(DatabaseProvider dbProvider, Graph graph, String
...
     private boolean indexFieldsExists(Vertex parentVertex, List<String> fields) {
         boolean contains = false;
-        if (parentVertex.property(Constants.INDEX_FIELDS).isPresent()) {
-            for (String field : fields) {
-                contains = fieldExists(parentVertex, field);
-                if(!contains)
-                    break;
-            }
+        for (String field : fields) {
+            contains = fieldExists(parentVertex, field);
```
indrajra: isn't this confusing? There is a list of fields and even if one of them doesn't exist, we try recreate index on all the list of fields. Why can't it be just the one that doesn't exist?

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
pritha-tarento: setProperty commit will happens with createIndex trx hence if createIndex fails the commit for setProperty also rolls backs.

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
pritha-tarento: Yes we can, Done.

java/registry/src/main/java/io/opensaber/registry/util/IndexHelper.java
```diff
@@ -0,0 +1,88 @@
...
+                return isIndexPresent;
+
+            List<String> indexUniqueFields = definition.getOsSchemaConfiguration().getUniqueIndexFields();
+            int uniqueIndexSize = (indexUniqueFields.size() == 0) ? 0
+                    : getNewFields(parentVertex, indexUniqueFields, true).size();
+            if (uniqueIndexSize > 0)
+                return isIndexPresent;
+
+        }
+        definitionIndexMap.put(defTitle, true);
```
indrajra: We must turn this true from outside...like in the next line of ensureIndexExists().

java/registry/src/main/java/io/opensaber/registry/util/IndexHelper.java
```diff
@@ -0,0 +1,88 @@
...
+     * 
+     * @param parentVertex
+     * @param definition
+     * @return
+     */
+    public boolean isIndexPresent(Vertex parentVertex, Definition definition) {
+        boolean isIndexPresent = false;
+        String defTitle = definition.getTitle();
+        logger.info("isIndexPresent flag for {}: {}", defTitle, definitionIndexMap.get(defTitle));
+        if (!definitionIndexMap.get(defTitle)) {
```
indrajra: Please check why can't this be simplified to just this -`return definitionIndexMap.getOrDefault(defTitle, false);`

java/registry/src/main/java/io/opensaber/registry/util/Indexer.java
```diff
@@ -0,0 +1,110 @@
...
+     * property
+     * 
+     * @param parentVertex
+     * @param values
+     * @param isUnique
+     */
+    private void updateIndices(Vertex parentVertex, List<String> values, boolean isUnique) {
+        String propertyName = isUnique ? Constants.UNIQUE_INDEX_FIELDS : Constants.INDEX_FIELDS;
+
+        if (values.size() > 0) {
```
indrajra: We must have this as a function in VertexWriter ideally...but its ok. I'm just leaving my thought here.

java/registry/src/main/java/io/opensaber/registry/util/EntityParenter.java
```diff
@@ -131,4 +140,60 @@ public Vertex getKnownParentVertex(String definition, String shardId) {
...
+     * 
+     * @param dbProvider
+     * @param parentVertex
+     * @param definition
+     */
+    public void ensureIndexExists(DatabaseProvider dbProvider, Vertex parentVertex, Definition definition) {
+        try{
+            if(!indexHelper.isIndexPresent(parentVertex, definition)){
+                asyncAddIndex(dbProvider, parentVertex, definition);
+            }
```
indrajra: After adding the index here, set the defintion->bool map to true.

PR #272, opened 2019-02-06
Issue #OS-188 added retry mechanism for signature and encryption service

Hari-stackroute:
Added retry mechanism for signature and encryption service.
configured retry-attempts and backoff values.

### DIFF
--- a/java/registry/src/main/resources/application.yml.sample
+++ b/java/registry/src/main/resources/application.yml.sample
@@ -148,8 +148,10 @@ validation:
   type: ${validation_type:json}
 
 retry:
+# The number of times an attempt must be made to reach to the service.
   maxAttempts: ${retry_maxAttempts:5}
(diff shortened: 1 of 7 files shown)


java/registry/src/main/resources/application.yml.sample
```diff
@@ -244,4 +249,9 @@ keycloak:
 validation:
   enabled: ${validation_enabled:true}
   type: ${validation_type:json}
+
+retry:
+  maxAttempts: ${signature_retry_maxAttempts:5}
```
indrajra: the env var must just be retry_maxAttempts, not signature_retry_maxAttempts. Same thing for the other var too.

Hari-stackroute: changed to retry_maxAttempts and retry_backoff_delay

PR #280, opened 2019-02-18
Os 191 Search Operators for String with limit & offset

pritha-tarento:
Search Api supports for String operators added.
Operators extended supports are: **contains, startsWith, endsWith, notContains, notstartsWith, notEndsWith**.
**or** operator : for querying multiple values enumerated
Also, provided **limit & offset** with request. limit/offset parameter are optional.

Sample request payload changes:
```
{"request": {
    "Teacher": {
      "teacherName": {"or": ["name1","name 2"] },
    },
  },
       "limit":10,
       "offset":2

}  
```
Here, **offset** is the start point (a number) and **limit** is the number of record return is present


java/registry/src/main/java/io/opensaber/registry/dao/SearchDaoImpl.java
```diff
@@ -27,23 +28,68 @@ public JsonNode search(Graph graphFromStore, SearchQuery searchQuery) {
 
 		GraphTraversalSource dbGraphTraversalSource = graphFromStore.traversal().clone();
 		List<Filter> filterList = searchQuery.getFilters();
-		GraphTraversal<Vertex, Vertex> resultGraphTraversal = dbGraphTraversalSource.clone().V().hasLabel(searchQuery.getRootLabel());
+		int offset = searchQuery.getOffset();
+		GraphTraversal<Vertex, Vertex> resultGraphTraversal = dbGraphTraversalSource.clone().V().hasLabel(searchQuery.getRootLabel())
+		        .range(offset, offset + 100).limit(searchQuery.getLimit()); //range max value is equal to default limit(100)
```
indrajra: This value of 100 cannot be hardcoded. It must come from a class member that has a link to application.yml.

indrajra: The limit and offset, why are they part of the the entity. It must be directly under "request", not under "request.Teacher".

pritha-tarento: ### done

pritha-tarento: done

PR #286, opened 2019-03-01
OS 195 - read document from elastic search

Hari-stackroute:
Added implementation for reading document from elastic-search
Found a issue with SignatureHelper, adding signatures as extranode. Taken care.
Added implemented methods for read document, update document, delete document in os elastic-search jar


java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -184,25 +183,39 @@ public String addEntity(String jsonString) throws Exception {
...
-    public JsonNode getEntity(String id, ReadConfigurator configurator) throws Exception {
-        DatabaseProvider dbProvider = shard.getDatabaseProvider();
-        IRegistryDao registryDao = new RegistryDaoImpl(dbProvider, definitionsManager, uuidPropertyName);
-        try (OSGraph osGraph = dbProvider.getOSGraph()) {
-            Graph graph = osGraph.getGraphStore();
-            Transaction tx = dbProvider.startTransaction(graph);
-            JsonNode result = registryDao.getEntity(graph, id, configurator);
+    public JsonNode getEntity(String id, String entityType, ReadConfigurator configurator) throws Exception {
+        JsonNode result = null;
+        Map<String, Object> response = elasticService.readEntity(entityType.toLowerCase(), id);
```
indrajra: If only the search provider is set to elastic search, we must do this. Otherwise, can't.

java/elastic-search/src/main/java/io/opensaber/elastic/ElasticServiceImpl.java
```diff
@@ -155,17 +161,37 @@ public RestStatus addEntity(String index, String entityId, Map<String, Object> i
...
     }
 
     @Override
-    public boolean updateEntity(String index, Map<String, Object> inputEntity, String osid) {
-        return false;
+    public RestStatus updateEntity(String index, Map<String, Object> inputEntity, String osid) {
+        logger.debug("updateEntity starts with index {} and entityId {}", index, osid);
+        UpdateResponse response = null;
+        try{
+            response = getClient(index).update(new UpdateRequest(index, searchType, osid), RequestOptions.DEFAULT);
```
indrajra: where is inputEntity getting used?

java/elastic-search/src/main/java/io/opensaber/elastic/ElasticServiceImpl.java
```diff
@@ -155,17 +161,37 @@ public RestStatus addEntity(String index, String entityId, Map<String, Object> i
...
 
     @Override
-    public void deleteEntity(String index, String osid) {
-
+    public RestStatus deleteEntity(String index, String osid) {
+        DeleteResponse response = null;
+        try{
+            response = getClient(index).delete(new DeleteRequest(index, searchType, osid), RequestOptions.DEFAULT);
+        } catch (IOException e) {
+            logger.error("Exception in delete a record from ElasticSearch", e);
```
indrajra: In all places you are catching the exception. What is the rest status response code indicate?
error (400) and message
or
success (200) and error message

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
Hari-stackroute: I have done the same way how search was implemented, created read for native-read and elasticsearch-read. While application loads the implementation is decided via config.

java/elastic-search/src/main/java/io/opensaber/elastic/ElasticServiceImpl.java
Hari-stackroute: modified the updated method, now it picks up input document for updation, but need to test.

java/elastic-search/src/main/java/io/opensaber/elastic/ElasticServiceImpl.java
Hari-stackroute: response status will be in codes, like 201- for created

PR #291, opened 2019-03-11
OS 182 cassandra support changes

Hari-stackroute:


java/registry/src/main/resources/application.yml.sample
```diff
@@ -53,7 +53,7 @@ database:
 
   # Providers available are NEO4J, SQLG, CASSANDRA, ORIENTDB, TINKERGRAPH (in-memory).
   # NOTE: Values given against 'shardId' must be unique
-  provider: ${database_provider:NEO4J}
+  provider: ${database_provider:CASSANDRA}
```
indrajra: Please do not change the default to cassandra. Let the example connection_uri be there and lets check in.

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -260,15 +260,15 @@ public void updateEntity(String id, String jsonString) throws Exception {
             logger.debug("After merge the payload is " + mergedNode.toString());
 
             // Re-sign, i.e., remove and add entity signature again
-            if (signatureEnabled) {
+            /*if (signatureEnabled) {
```
indrajra: Investigation required if the value of "signatureFor" causes any problem.

java/registry/src/main/resources/application.yml.sample
Hari-stackroute: for the connection_uri changes are done, actually database_provider should be changed to cassandra. Is my understanding right, on your ask?

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
Hari-stackroute: Faced a problem with JanusGraph api, vertex property cannot be labeled to any edge. Did modifications accordingly.

PR #295, opened 2019-03-22
OS 201 Retry to elascticsearch read and search api's and misc changes

Hari-stackroute:
Added retry mechanism for read and search api's for elastic search
Added a flag for add api for check and commit(helps for testing)

What missed:
Need to do same kind of retry operation for update and add api's too, since these are performed through actor unable to proceed for as of now.

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/service/ElasticReadService.java
+++ b/java/registry/src/main/java/io/opensaber/registry/service/ElasticReadService.java
@@ -13,9 +13,12 @@ import io.opensaber.registry.middleware.util.Constants;
 import io.opensaber.registry.middleware.util.DateUtil;
 import io.opensaber.registry.middleware.util.JSONUtil;
 import io.opensaber.registry.util.ReadConfigurator;
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
@@ -26,6 +29,8 @@ import org.springframework.stereotype.Component;
  */
 @Component
 public class ElasticReadService implements IReadService {
+
+    private static Logger logger = LoggerFactory.getLogger(ElasticReadService.class);
 
     @Autowired
     private IElasticService elasticService;
@@ -51,7 +56,12 @@ public class ElasticReadService implements IReadService {
     public JsonNode getEntity(String id, String entityType, ReadConfigurator configurator) throws Exception {
         JsonNode result = null;
         AuditRecord auditRecord = null;
-        Map<String, Object> response = elasticService.readEntity(entityType.toLowerCase(), id);
+        Map<String, Object> response = null;
+        try{
+            response = elasticService.readEntity(entityType.toLowerCase(), id);
+        } catch (IOException e) {
+            logger.error("Exception in reading a record to ElasticSearch", e);
+        }
         if(response == null) {
             throw new RecordNotFoundException("Record with " +id+ " not found");
         }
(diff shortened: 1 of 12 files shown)


java/registry/src/main/java/io/opensaber/registry/service/ElasticReadService.java
```diff
@@ -51,7 +56,12 @@
     public JsonNode getEntity(String id, String entityType, ReadConfigurator configurator) throws Exception {
         JsonNode result = null;
         AuditRecord auditRecord = null;
-        Map<String, Object> response = elasticService.readEntity(entityType.toLowerCase(), id);
+        Map<String, Object> response = null;
+        try{
+            response = elasticService.readEntity(entityType.toLowerCase(), id);
+        } catch (IOException e) {
```
indrajra: Couldn't RecordNotFound exception be thrown in the catch block?

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -93,6 +93,9 @@
     @Value("${persistence.enabled}")
```
indrajra: Default this to false and remove from application.yml. Persistence.enabled and Persistence.commit.enabled are internal dev-only flags and it is best we dont expose these.

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -93,6 +93,9 @@
...
     private boolean persistenceEnabled;
 
+    @Value("${commit.enabled:false}")
```
indrajra: persistence.commit.enabled

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -200,57 +203,22 @@ public String addEntity(String jsonString) throws Exception {
...
                 tx = dbProvider.startTransaction(graph);
                 entityId = registryDao.addEntity(graph, rootNode);
-                shard.getDatabaseProvider().commitTransaction(graph, tx);
-                dbProvider.commitTransaction(graph, tx);
-
+                if (commitEnabled) {
+                    dbProvider.commitTransaction(graph, tx);
+                }
+            } finally {
+                tx.close();
```
indrajra: tx closure's are supposted to be handled by osGraph scope. Let's not write explicit.

java/registry/src/main/java/io/opensaber/registry/service/ElasticReadService.java
Hari-stackroute: Here after certain retries, if the service is still unavailable or some exception happens in the elastic-search it will throw IOException, so here we will be logging the IOException and after that response anyways will be null, so will be throwing record not found too.

In general scenario, one can ask for invalid record and we throwing record not found.

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
Hari-stackroute: Here we are not opening the transaction with try(), when we are testing without committing for perf. the first request which takes a transaction will not be closed since it is not committed, so when we are triggering a second request application getting hanged. so i am externally closing it.

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
Hari-stackroute: Ok

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
Hari-stackroute: ok

PR #300, opened 2019-04-04
Os 209 - osid not getting created 

pritha-tarento:
osid was not getting created for all Schema definition(like Person, Vehicle, etc..)

Changes done:
Introduced a IndexFields pojo to hold all types of indices.
This gets populated one time while application starts.

Some more changes done to handle few boundary conditions IndexHelper

### DIFF
--- a/java/registry/src/main/java/io/opensaber/registry/util/EntityParenter.java
+++ b/java/registry/src/main/java/io/opensaber/registry/util/EntityParenter.java
@@ -4,6 +4,7 @@ import io.opensaber.registry.dao.VertexWriter;
 import io.opensaber.registry.middleware.util.Constants;
 import io.opensaber.registry.model.DBConnectionInfo;
 import io.opensaber.registry.model.DBConnectionInfoMgr;
+import io.opensaber.registry.model.IndexFields;
 import io.opensaber.registry.sink.DBProviderFactory;
 import io.opensaber.registry.sink.DatabaseProvider;
 import io.opensaber.registry.sink.OSGraph;
@@ -48,6 +49,11 @@ public class EntityParenter {
      * Holds information about a shard and a list of definitionParents
      */
     private HashMap<String, ShardParentInfoList> shardParentMap = new HashMap<>();
+    /**
+     * Holds information for all definitions and it's indices 
+     */
+    private Map<String, IndexFields> definitionIndexFields = new ConcurrentHashMap<String, IndexFields>();
+
 
     @Autowired
     public EntityParenter(DefinitionsManager definitionsManager, DBConnectionInfoMgr dbConnectionInfoMgr) {
@@ -63,6 +69,7 @@ public class EntityParenter {
      */
     public void loadDefinitionIndex() {
         Map<String, Boolean> indexMap = new ConcurrentHashMap<String, Boolean>();
+        
 
         for (Map.Entry<String, ShardParentInfoList> entry : shardParentMap.entrySet()) {
             String shardId = entry.getKey();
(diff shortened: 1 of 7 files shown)


java/registry/src/main/java/io/opensaber/registry/util/EntityParenter.java
```diff
@@ -71,19 +78,32 @@ public void loadDefinitionIndex() {
                 Definition definition = definitionsManager.getDefinition(shardParentInfo.getName());
                 Vertex parentVertex = shardParentInfo.getVertex();
                 List<String> indexFields = definition.getOsSchemaConfiguration().getIndexFields();
+                if (!indexFields.contains(uuidPropertyName)) {
+                    indexFields.add(uuidPropertyName); // adds default field
```
indrajra: Are you sure adding here persists for later use?

java/registry/src/main/java/io/opensaber/registry/util/IndexHelper.java
```diff
@@ -81,9 +84,15 @@ public boolean isIndexPresent(Definition definition, String shardId) {
     public static List<String> getCompositeIndexFields(List<String> fields) {
         if (fields.size() > 0) {
             StringBuilder list = new StringBuilder(String.join(",", fields));
-            String subString = StringUtils.substringBetween(list.toString(), "(", ")").replaceAll("'", "");
-            String[] commaSeparatedArr = subString.split("\\s*,\\s*");
-            List<String> result = Arrays.stream(commaSeparatedArr).collect(Collectors.toList());
+            List<String> result;
```
indrajra: I think we must iterate and do this substring comparison only if we find ( or ) in the first place. Am not sure if this code helps.

java/registry/src/main/java/io/opensaber/registry/util/Indexer.java
```diff
@@ -69,9 +69,12 @@ public void setCompositeIndexFields(List<String> compositeIndexFields) {
     public void createIndex(Graph graph, String label) {
 
         if (label != null && !label.isEmpty()) {
-            databaseProvider.createUniqueIndex(graph, label, indexUniqueFields);
-            databaseProvider.createIndex(graph, label, singleIndexFields);
-            databaseProvider.createCompositeIndex(graph, label, compositeIndexFields);
+			if (indexUniqueFields.size() > 0)
```
indrajra: Was this necessary? I think otherwise from code. We always have an empty List and create?Index() functions does an iteration. Hence my thought.

java/registry/src/main/java/io/opensaber/registry/util/EntityParenter.java
pritha-tarento: Yes.

java/registry/src/main/java/io/opensaber/registry/util/Indexer.java
pritha-tarento: Necessary because if any of this fields size =0, while transaction throws error could not create

java/registry/src/main/java/io/opensaber/registry/util/IndexHelper.java
pritha-tarento: Done

java/registry/src/main/java/io/opensaber/registry/util/IndexHelper.java
```diff
@@ -105,22 +109,25 @@ public boolean isIndexPresent(Definition definition, String shardId) {
...
-            List<String> result = Arrays.stream(commaSeparatedArr).collect(Collectors.toList());
-            return result;
-        } else {
-            return new ArrayList<>();
+		List<String> result = new ArrayList<String>();
+		if (fields.size() > 0) {
 
-        }
+			String[] commaSeparatedArr = new String[fields.size()];
+			for (int i = 0; i < fields.size(); i++) {
```
indrajra: I think this must be simple like this
`
for (int i = 0; i < fields.size(); i++) {
				String field = fields.get(i);
				boolean containsCompositeValues = (field.indexOf("(") == 0)
						&& (field.indexOf(")") == field.length() - 1);
				if (!containsCompositeValues) {
                                     result.add(field);
                                 }
}
`

java/registry/src/main/java/io/opensaber/registry/util/Indexer.java
indrajra: Can't you add once in the implementation?
`private void ensureIndex(VertexLabel vertexLabel, IndexType indexType, List<PropertyColumn> properties) {
if (properties.size > 0) {
        Index index = vertexLabel.ensureIndexExists(indexType, properties);
        logger.info(indexType + "index created for " + vertexLabel.getLabel() + " - " + index.getName());
}
    }
`

pritha-tarento: Done

java/registry/src/main/java/io/opensaber/registry/util/IndexHelper.java
pritha-tarento: Done

java/registry/src/main/java/io/opensaber/registry/util/EntityParenter.java
indrajra: It doesnt, but you have moved this addition to a new class (IMO, un-needed) and so works.

PR #303, opened 2019-04-08
Os 210 Logback rotations enabled

pritha-tarento:
Logback rotations enabled on size limit 30Mb
MaxSize of log file could be 30Mb. If exceeds rotation takes place.
This rotation is done for perf logs, audit logs, app logs.

### DIFF
--- a/java/registry/src/main/resources/logback.xml
+++ b/java/registry/src/main/resources/logback.xml
@@ -8,35 +8,18 @@
 
     <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>logs/app.log</file>
-        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
             <!-- daily rollover -->
-            <fileNamePattern>logFile.%d{yyyy-MM-dd}.log</fileNamePattern>
-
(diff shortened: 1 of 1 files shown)


java/registry/src/main/resources/logback.xml
```diff
@@ -55,15 +42,14 @@
...
 
     <appender name="AuditFileAppender" class="ch.qos.logback.core.rolling.RollingFileAppender">
-        <file>logs/audit.log</file>
-        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
-            <fileNamePattern>logs/audit-%d{yyyy-MM-dd}.log</fileNamePattern>
-            <maxHistory>30</maxHistory>
-            <totalSizeCap>5GB</totalSizeCap>
+        <file>audit_logs/audit.log</file>
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+            <fileNamePattern>audit_logs/audit-%d{yyyy-MM-dd}_%d{HH-mm-ss}.%i.log</fileNamePattern>
```
indrajra: audit has hyphen, perf has dot, and logFile has dot - please use underscore as uniform separator.
Also, rename the logFile to app...All the file names will be one of the following - app, perf and audit. They may have date_time_index suffix if they are rotated. As simple as that.

pritha-tarento: Done

PR #307, opened 2019-04-25
OS#224 Create a sample UI for Person schema

indrajra:
Enabled CORS in the app so that adopters can quickly start integration. This was encountered even during our devcon days and we had enabled it then by hard-code. Now, I've externalised the origins to the configuration. 
Explicitly added port 8080 to application configuration. Thought this was already added, but find it missing.
Used Alpaca - a json-schema form generator to generate a dynamic form for the Person schema.
Turned off perf - default option must be that.
If the database shard is set to "none", the class name provided must be ignored. Presently, the class is attempted to instantiate and throws an exception.
The class name for shard advisor property was not specified properly and so was failing. 
Note: The last two items were part of OS-222, but being taken as the files are same.

Testing:
1. Able to do 'required' validations.
2. Able to add a record to the database from the UI.

As part of this PR, the following is not done:
1. Other than Add operations, we need some UI buttons and html files to showcase.
2. Validations to be done at UI level with ajv {another json schema validator}
3. Other cosmetic things like layout - this can be done using alpaca. A sample could be provided in our repo.

### DIFF
--- a/ui-sample/schema/Person.json
+++ b/ui-sample/schema/Person.json
@@ -1,0 +1,56 @@
+{
+  "$id": "#/properties/Person",
+  "type": "object",
+  "title": "The Person Schema",
+  "required": [
+    "nationalIdentifier",
+    "name",
+    "gender",
+    "dob"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "photo": {
+      "type": "string",
+      "contentEncoding": "base64",
+      "mediaType": "image/png",
+      "title": "photo"
+    },
+    "nationalIdentifier": {
+      "$id": "#/properties/nationalIdentifier",
+      "type": "string",
+      "$comment": "Nationality",
+      "title": "National identifier"
+    },
+    "name": {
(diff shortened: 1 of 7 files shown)


ui-sample/schema/Person.json
```diff
@@ -0,0 +1,56 @@
...
+  "title": "The Person Schema",
+  "required": [
+    "nationalIdentifier",
+    "name",
+    "gender",
+    "dob"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "photo": {
```
Hari-stackroute: photo is not there in the required fields and additionalproperties mentioned as false, it means photo field should not allowed is in't?

indrajra: It just means photo is a valid field, but is not mandatory to supply. Any other field, given would be disallowed because "additionalProperties" is set to false.

Hari-stackroute: ok i understood as viceversa.

PR #310, opened 2019-05-09
OS 227 misc issues during load testing for user data model

Hari-stackroute:
Create an actor configuration copy of what was used in testing (load)
Search provider case mismatches fix
The actor factory was not creating actors in the cloud boxes (they do in laptops).


java/registry/src/main/resources/application.yml.sample
```diff
@@ -32,15 +32,15 @@ registry:
     base: ${registry_system_base:http://localhost:8080/opensaber/}
   rootEntity:
         type: ${registry_rootentity_type:Teacher}
- 
-# This property is to be used for search request/query 
-# offset: search result starts from value provided as default        
-# limits: search result to value provided as default 
+
```
indrajra: This file changes could be reverted except for picking up env var, perhaps.

Hari-stackroute: changes done, merged with 2.0.3

PR #311, opened 2019-05-17
OS 228 Read data from DB when ES down/fails

Hari-stackroute:
Read data from backend DB when unable to read from ES(read immediate after write failing sometimes when threads are high, found in load test)
During search osid can be passed to filters with shard-id, this is failing in previous releases.
Created user_vt for user load testing.


java/registry/src/main/java/io/opensaber/registry/service/ElasticReadService.java
```diff
@@ -44,6 +44,8 @@
     private IAuditService auditService;
     @Autowired
     private APIMessage apiMessage;
+    @Autowired
```
indrajra: ElasticReadSearch must not know about NativeReadSearch. There should not be any changes in Elastic services, but only in the service/control layers above.

java/registry/src/main/java/io/opensaber/registry/service/ISearchService.java
```diff
@@ -126,6 +133,18 @@ default void addToFilterList(String path, JsonNode inputQueryNode, List<Filter>
         return rangeValues;
     }
 
+    default JsonNode replaceKeyValue(JsonNode jsonNode, String uuidPropertyName) {
+        ObjectNode osidKeyValue = null;
+        if(jsonNode.findPath(uuidPropertyName).isObject()) {
+            osidKeyValue = (ObjectNode)jsonNode.get("filters").get(uuidPropertyName);
+            String nodeKey = osidKeyValue.fields().next().getKey();
+            RecordIdentifier recordID = RecordIdentifier.parse(osidKeyValue.get(nodeKey).asText());
```
indrajra: what about those that do not parse? Exception not handled here, I think.
Is the if condition really necessary? (Un-necessary ifs will cause less code-coverage numbers, mainly branch. Please avoid them.)

Hari-stackroute: Here we are checking if the filters only contains the uuidproperty, else we are not allowed to parse. And the check here is necessary, can't slip.

PR #335, opened 2019-11-26
Pagination: Removed extra search call for getting totalItems and added package ngx-pagination

aishwa8141:


indrajra:
We worked on this in a different way. Closing this.

PR #341, opened 2019-12-10
Created  Notification module, Http util, registry service, Keycloak helper.js

aishwa8141:
Notification Module 
Http utils contain post and get methods
Registry service is used to connect open saber
Keycloakhelper.js helps to get access token and register employee and getUserByRoles


ui-sample/util-service/keycloakHelper.js
```diff
@@ -0,0 +1,93 @@
+
+var keyCloakAuthUtils = require('keycloak-auth-utils');
+let request = require('request')
+const httpUtil = require('./httpUtils.js');
+const realmName = process.env.keycloak_realmName || "PartnerRegistry"
+const keyCloakHost = process.env.keycloak_url || "http://localhost:8080/auth/admin/realms/" + realmName;
+
+let keyCloak_config = {
```
indrajra: isn't this same as in keycloak.json

ui-sample/util-service/notification.js
```diff
@@ -0,0 +1,101 @@
...
+                        }
+                    }
+                });
+            }
+
+        }
+    });
+}
+
+function getUser(value, tokenDetails, callback) {
```
indrajra: any interaction with other modules must be layered above. not here.

ui-sample/util-service/notifyRulesSet.json
```diff
@@ -0,0 +1,16 @@
+{
+        "create":{
```
indrajra: add a key '$comment' to explain what is this.
Every key must have uniform set of key-values. We must discuss on this.

indrajra: We discussed and would move this differently.

PR #356, opened 2020-01-14
Issue #OS-241 adding audit schema creation feature

sknirmalkar89:


sknirmalkar89:
Change is not required as we are expecting adopter to create a audit template of a subject and put inside schema folder for audit.

PR #372, opened 2020-02-06
Search templates renamed with uuid

aishwa8141:
1)added subject to the Github Id (task 6)
2)search templates name must be uuid ( task 7)
3) isOnboarded flag should true if isActive is true  and endDate is null( task 9)
4)Removed check for required fields in update component (task10)
5)task 7 (updation in NER should reflect in EPR and notify)


scripts/index.js
```diff
@@ -113,7 +113,7 @@ var populate_add_tasks = function (tasks, entityType, static_payload, arrDynamic
                 dataPortion[field] = myArr
             }
             if (field === 'isActive') {
-                if (dataPortion[field] === 'Yes') {
+                if (dataPortion[field] === 'Yes' && !dataPortion['endDate']) {
```
indrajra: This condition is not a logical AND.
isOnboarded = isActive || (endDate != null)

aishwa8141: done

PR #375, opened 2020-02-24
OS-250 Removing JSONArray dependency in ArrayHelper.java and Fixing S…

reshmi-nair:
…ignature issue with viewTemplate

### DIFF
--- a/java/view-templates/src/main/java/io/opensaber/views/ViewTransformer.java
+++ b/java/view-templates/src/main/java/io/opensaber/views/ViewTransformer.java
@@ -1,6 +1,7 @@
 package io.opensaber.views;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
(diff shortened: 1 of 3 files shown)


java/view-templates/src/main/java/io/opensaber/views/ViewTransformer.java
```diff
@@ -34,15 +35,18 @@ public JsonNode transform(ViewTemplate viewTemplate, JsonNode node) throws Excep
 				ArrayNode resultArray = JsonNodeFactory.instance.arrayNode();
 
 				for (int i = 0; i < nodeAttrs.size(); i++) {
+					
 					JsonNode tNode = tranformNode(viewTemplate, nodeAttrs.get(i));
+					appendSignatures(viewTemplate, nodeAttrs.get(i),tNode);
```
indrajra: Move appendSignatures inside transformNode() .

java/view-templates/src/main/java/io/opensaber/views/ViewTransformer.java
```diff
@@ -97,4 +101,32 @@ private JsonNode tranformNode(ViewTemplate viewTemplate, JsonNode nodeAttrs) thr
         }
         return result;
     }
-}
+    
+    private JsonNode appendSignatures(ViewTemplate viewTemplate, JsonNode nodeAttrs, JsonNode resultNode) throws Exception {
+    	if(nodeAttrs.get("signatures")!=null) {
+			
+			ArrayNode sigArray = new ObjectMapper().createArrayNode();
```
indrajra: should we iterate or just would be able to say like resultNode['signatures'] = nodeAttrs.get['signatures'] ?

reshmi-nair: iteration is to specifically find out the signatures of fields specified in viewtemplate.

java/view-templates/src/main/java/io/opensaber/views/ViewTransformer.java
reshmi-nair: this is done.

PR #384, opened 2020-03-03
Issue#EPR-15:Added form template for x-owner role

aishwa8141:
1)Added check to show the profile page for x-owner in the header component 
2)added cancel button in the update profile page


ui-sample/util-service/templates/X-OwnerFormTemplate.json
```diff
@@ -0,0 +1,321 @@
...
+                },
+                "index": 7
+            },
+            {
+                "code": "role",
+                "dataType": "text",
+                "name": "role",
+                "label": "EkStep role",
+                "description": "EkStep role",
+                "editable": true,
```
indrajra: For an x-owner, none of the fields are editable. Even if we don't expose the buttons, lets also fix this.

ui-sample/util-service/templates/X-OwnerFormTemplate.json
```diff
@@ -0,0 +1,321 @@
...
+                "placeholder": "Project name",
+                "inputType": "select",
+                "required": false,
+                "displayProperty": "Editable",
+                "visible": true,
+                "renderingHints": {
+                    "semanticColumnWidth": "four"
+                },
+                "range": [
```
indrajra: Consider making all of this range blank, since it is not going to be edited.

aishwa8141: done

ui-sample/util-service/templates/X-OwnerFormTemplate.json
aishwa8141: done

PR #390, opened 2020-03-05
Issue#EPR-18: Add ability to download users in bulk as csv 

aishwa8141:
1) used npm library json2csv to convert json to csv [https://www.npmjs.com/package/json2csv](url)
2)To dowload and save csv file used npm package file-saver [https://www.npmjs.com/package/file-saver](url)
2)created service called DowloadCsvFile


ui-sample/angular-ui/src/app/components/admin-page/admin-page.component.html
```diff
@@ -23,6 +23,8 @@
           [popupPlacement]="'bottom right'">
           <i class="icon {{buttonIcon}}"></i>
         </div>
+        <a  class="ui button primary end"  *ngIf='permissionService.permissionAvailable' appPermission
+        [permission]="downloadCsvRoles" (click)="downloadFile()"  [ngClass]="{'loading' : loader}">Download CSV</a>
```
indrajra: Call it as Download as much as possible. We may download as csv, pdf or other formats later. So, if you think likewise, you will automatically call roles as 'downloadRole'.

ui-sample/angular-ui/src/app/services/download/download-csvfile.service.ts
```diff
@@ -0,0 +1,29 @@
...
+
+
+  Json2csvParser = json2csv.Parser;
+  constructor() {
+
+  }
+  public downloadCsvFile(data: any, filename?: string) {
+    let csvData = this.convertJsonToCSV(data);
+    let file = new Blob([csvData], { type: 'text/csv;charset=utf-8' });
+    saveAs(file, "Employees.csv");
```
indrajra: How will this be handling multiple connections? Guess you must add a timestamp to say the least.
And in the constructor remove the ones older than 5 min or so, so that we keep the container size under control.

ui-sample/angular-ui/src/app/components/admin-page/admin-page.component.html
aishwa8141: done

ui-sample/angular-ui/src/app/services/download/download-csvfile.service.ts
indrajra: Sorry, this is in ui. So no issues.

PR #398, opened 2020-03-16
Removed /admin from the keycloak add user api url

aishwa8141:
and removed check isActive while registered new user to the keycloak


aishwa8141:
refer PR 400

PR #399, opened 2020-03-16
EPR-41 Display QrCode in Employee Profile page

reshmi-nair:
Display QrCode in Employee Profile page


ui-sample/util-service/index.js
```diff
@@ -84,6 +85,25 @@ app.theApp.post("/offboard/user", (req, res, next) => {
         }
     });
 });
+app.theApp.post("/profile/qrImage", (req, res, next) => {
```
indrajra: pass the user id or let this be an implicit call (meaning when the read profile happens, can we have this populated from this layer for the UI? )

Sending qr code to this API is not an idea. Everyone can send and get a QR. Recall we discussed we can put some signature within QR to say that this belongs to EkStep (later...).

ui-sample/util-service/index.js
```diff
@@ -85,6 +87,42 @@ app.theApp.post("/offboard/user", (req, res, next) => {
...
+      }
+      QRCode.toDataURL(req.query.qrCode, opts, function (err, url) {
+        if (err) throw err
+
+        if(url.indexOf('base64') != -1) {
+            var buffer = Buffer.from(url.replace(/^data:image\/png;base64,/, ""), 'base64');
+            Jimp.read(buffer, (err, image) => {
+            if (err) throw err;
+            else {
+                Jimp.loadFont(Jimp.FONT_SANS_16_BLACK).then(function(font) {
```
indrajra: If we change the font and adjust height it could appear beautiful I think. Also, what about margin and color?

PR #400, opened 2020-03-16
ISSUE#EPR-43:Fetch and update employee code during registration

aishwa8141:
Keycloak add user API URL updated
from       url: this.keyCloakHost + "/auth/admin/realms/" + this.realmName + "/users"
to            url: this.keyCloakHost + "/auth/realms/" + this.realmName + "/users/add"


ui-sample/util-service/index.js
```diff
@@ -217,21 +244,16 @@ const getTokenDetails = (req, callback) => {
...
+    if (res.statusCode == 200) {        
+        req.body.request[entityType]['kcid'] = res.body.id
         req.body.request[entityType]['isOnboarded'] = req.body.request[entityType].isActive;
+        req.body.request[entityType]['empCode'] = employeeCode.prefix  + employeeCode.nextCode;
         registryService.addRecord(req, function (err, res) {
-            if (res.statusCode == 200 && res.params.status == 'SUCCESSFUL') {
-                logger.info("record successfully added to registry")
+            if (res.statusCode == 200 && res.body.params.status == 'SUCCESSFUL') {
+                logger.info("record successfully added to registry");
+                updateEmployeeCode(employeeCode, req.headers)
```
indrajra: before calling addRecordToRegistry, why don't we call this in the series?...I think waterfall is not the correct one here. perhaps series is befitting.

PR #404, opened 2020-03-19
EPR-36 Solving userid null issue in OS audit service

reshmi-nair:
When userid is null, audit was not getting added. Fixed this issue in OS audit service


indrajra:
This is not needed. Fix was found another way.

PR #413, opened 2020-04-13
EPR-90 Fix Read api for array of object of size 1 and ensure update objects work

sknirmalkar89:


java/registry/src/main/java/io/opensaber/registry/dao/VertexWriter.java
```diff
@@ -162,18 +174,31 @@ public void createArrayNode(Vertex vertex, String entryKey, ArrayNode arrayNode)
...
+    
+    /**
+     *  Add single entity node to the database
+     *  
+     * @param vertex
+     * @param label
+     * @param entryValue
+     * @return 
+     */
+    public Vertex writeSingleEntityNode(Vertex vertex, String label, JsonNode entryValue) {
```
indrajra, line 188: Why this refactoring? Void methods if starting to return a value doesn't impact any. Could find the only intention is the return the vertex just added.

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
```diff
@@ -327,26 +331,39 @@ private void doUpdateArray(Shard shard, Graph graph, IRegistryDao registryDao, V
...
+                } else {
+                    // New item got added.
+                    Vertex newItem = vertexWriter.writeSingleEntityNode(blankArrVertex, vr.getInternalType(blankArrVertex), item);
+                    updatedUuids.add(shard.getDatabaseProvider().getId(newItem));
                 }
             }
         }
+        
+        //Update the array_node with list of update uuids
+        vertexWriter = new VertexWriter(graph, shard.getDatabaseProvider() , uuidPropertyName);
```
indrajra, line 344: is this really needed? Can't the same object be used.

java/registry/src/main/java/io/opensaber/registry/dao/VertexWriter.java
```diff
@@ -96,6 +96,18 @@ public void updateParentIndexProperty(Vertex parentVertex, String propertyName,
...
+    
+    /**
+     * Update single property of the vertex
+     * 
+     * @param vertex
+     * @param propertyName
+     * @param value
+     */
+    
+    public void updateVertexProperty(Vertex vertex, String propertyName, String value) {
```
indrajra, line 108: This function is probably must be named as 'updateArrayNode' and these lines come in here..
`String propertyName = RefLabelHelper.getLabel(vr.getInternalType(blankArrVertex), uuidPropertyName);
        vertexWriter.updateVertexProperty(blankArrVertex, propertyName,ArrayHelper.formatToString(new ArrayList<>(updatedUuids))); `

The property is sent freely in the VertexWriter class and just wrapping them with 'update' name isn't gelling well.
It could be setPropery() but that's overkill. If you can bring in blank node operations here, the change is meaningful.

java/registry/src/main/java/io/opensaber/registry/dao/VertexWriter.java
sknirmalkar89: The intention was not to affect the current signature of the function while refactoring since we need to return the vertex after adding, though it doesn't affect even when we return. But the purpose of the signature of the method to use the return type.

java/registry/src/main/java/io/opensaber/registry/service/impl/RegistryServiceImpl.java
sknirmalkar89: this can be removed, I missed to remove it.

java/registry/src/main/java/io/opensaber/registry/dao/VertexWriter.java
indrajra: :-) not convincing, though I appreciate the care. Keep up the care!

PR #449, opened 2021-04-11
invite endpoint

saisathvik:
* creates an entry for the entity
* creates Keycloak user based on subjectJsonPath


java/registry/src/main/java/io/opensaber/registry/util/KeycloakAdminUtil.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Collections;
+import javax.ws.rs.core.Response;
+
+
+@Component
+public class KeycloakAdminUtil {
+    private static final Logger logger = LoggerFactory.getLogger(KeycloakAdminUtil.class);
+
+    private static final String realm = "divoc";
+    private static final String clientSecret = "daa3e6b4-7be0-4cc5-9854-14ea15f74cae";
```
dileepbapat, line 25: auto wire from config

